Issue #218, opened 2014-09-24
Use `.js` in module names


guybedford:
I believe this is an idea that got quite a bit of support in the last TC39, as mentioned in https://github.com/google/traceur-compiler/issues/1221.

Pending feedback from this current meeting, it may be implementation ready soon. It is only an adjustment to the System loader hooks.

PR #219, opened 2014-09-24
disable auto js extensions

guybedford:
Here's a prototype for https://github.com/ModuleLoader/es6-module-loader/issues/218.

* In normalize, we normalize relative to the parentName, and when none is provided we normalize relative to the baseURL.
* Paths are applied in normalize, before normalization, only on names that do not start with "scheme:", "./", "../", "/".
* Locate does a final normalization relative to the baseURL

All tests passing.

Feedback welcome @caridy.


guybedford:
This has been replaced by https://github.com/ModuleLoader/es6-module-loader/pull/317.

Issue #167, opened 2014-06-25
use a test runner


calvinmetcalf:
mocha works in both browsers and node so might be a good one to use, tape with testling might be a good choice as well

guybedford:
Massive thanks to @douglasduteil.

PR #250, opened 2014-11-20
chore(test): use mocha

douglasduteil:
Following #167 and #247

~~Here is a start with half of the tests.
Wait for it :)~~


guybedford:
This has been replaced with #274 I take it.

PR #274, opened 2014-12-30
refacto(test): karmify the tests

douglasduteil:
Hi @guybedford 

Following #250 
Here is a _karma-fication_ of the tests. [Travis result here](https://travis-ci.org/douglasduteil/es6-module-loader/builds/45437849)

- `npm test` : run the tests without the polyfill
- `npm test:polyfill` : run the tests with the polyfill
- `npm test:perf` : run the benchmarks

Notes: 
- A lot of `karma-*.conf.js` files just to keep things simple. A common script can be made in the future.
- The `script[type="module"]` test is still unsolved. Don't really know how to _karmify_ it properly. Suggestions are welcome.
- The benchmarks results aren't fixed. I means it's normal depending of where it's running, but really Travis has some pretty low results... Tell me if it's an issue for you. We can just remove the benchmarks runner on it if you want.

Closes #250


guybedford:
@douglasduteil this looks really great.

Two big questions - what about running the tests in IE8 and NodeJS?

douglasduteil:
> what about running the tests in IE8

Yes today the tests fail on IE8 due to Chai. I guess I will make add a work around with [karma-expect](https://www.npmjs.com/package/karma-expect).

> what about running the tests in NodeJS

Supa simple all the test factory is "node-made" so it will work without toutching it by adding some global in Node like `expect`

guybedford:
@douglasduteil that sounds amazing, thanks. I will happily merge once these are in.

douglasduteil:
It's is good to you I'll rebase the above commit to one only commit of test refactorization in another branch and officialize some npm scripts like :

- `npm run test:node` will use node to  to run the tests 
- `npm run test:browsers` will use karma to run the tests 
- `npm run test:browsers+polyfill` will use karma to run the tests with the polyfill
- `npm run test:browsers:sl` will use karma and saucelabs to run the tests in various browsers [*].
- `npm run test:browsers:sl+polyfill`  will use karma and saucelabs to run the tests in various browsers with the polyfill
- `npm run test:browsers:perf` will use karma to run benchmarks.
- `npm test` run `npm run test:node && npm run test:browsers`


[*] It would be cool if you can give me the list of browsers you want to test on. [I listed most of the saucelab browsers here](https://github.com/douglasduteil/gesalakacula#sample)

guybedford:
The code is carefully written with minimal browser requirements, so the IE versions are the main thing. Would be great to test a bigger variety though too - whatever you think is best.

Note also that the tests only run up to https://github.com/ModuleLoader/es6-module-loader/blob/master/test/test.js#L173 in IE8 actually, so perhaps we can even ignore IE8, as I normally catch the major coverage here through the SystemJS tests themselves.

guybedford:
Having some basic tests for IE8 is certainly important though to catch the basic compatibility.

douglasduteil:
Alright

douglasduteil:
Hum...

douglasduteil:
The `traceur@0.0.79` is failing on IE8... normal ?

guybedford:
It looks like this has been updated with https://github.com/ModuleLoader/es6-module-loader/pull/284.

Thanks @douglasduteil for the hard work on this, just let me know if there is anything else I can do to help get this towards a merge.

PR #279, opened 2015-01-03
clean up npm dependencies

pkozlowski-opensource:
grunt-contrib-uglify was present in both dependencies and devDependencies
which probably wasn't intentional


guybedford:
Thanks.

PR #284, opened 2015-01-22
Testify against the machine !

douglasduteil:
#### Here comes the big changes on the tests.

I'm end up with 49 tests on Node and with 50 per browser.
@guybedford Is it the right count ?

Result : [![Build Status][travis-image]][travis-url]  [![Sauce Test Status][saucelabs-image]][saucelabs-url]

[![Sauce Test Status][saucelabs-browser-matrix-image]][saucelabs-url]

[saucelabs-url]: https://saucelabs.com/u/de-es6-module-loader
[saucelabs-image]: https://saucelabs.com/buildstatus/de-es6-module-loader
[saucelabs-browser-matrix-image]: https://saucelabs.com/browser-matrix/de-es6-module-loader.svg
 
[travis-url]: https://travis-ci.org/douglasduteil/es6-module-loader
[travis-image]: https://travis-ci.org/douglasduteil/es6-module-loader.svg?branch=testify-against-the-machine

PR #290, opened 2015-01-25
test: fix ie9 console.log error

douglasduteil:
On SauceLabs the instantiated IE browser can be available without developer tools opened.
And for IE 8-9 it means no "console" object defined on window...


guybedford:
Why would we want to set `console.log` to alert? Surely we should use something non-blocking?

guybedford:
@douglasduteil shouldn't this build failure be reporting the whole build as a failure https://travis-ci.org/ModuleLoader/es6-module-loader/jobs/48320315#L473 ?

douglasduteil:
> Why would we want to set console.log to alert? Surely we should use something non-blocking?

K. I set to alert to make sure IE was responding to the logs in the tests...
Sry for the tests I forgot to test on node yesterday...

Anyway. What you think about using the [`window.dump`](https://developer.mozilla.org/en-US/docs/Web/API/window.dump) or an empty function if no console.log exist ?

guybedford:
An empty function seems fine.

Let me know about that other point when you can too.

douglasduteil:
:up:

douglasduteil:
Rocks here : https://travis-ci.org/douglasduteil/es6-module-loader/jobs/48323469#L401

guybedford:
Nice! I'd still be very interested to hear why the issue above wasn't returning a full build failure though? How could a failing test result in a passing build?

douglasduteil:
Ha !  I forgot to say it.

If you're looking at the [Travis output of this repo](https://travis-ci.org/ModuleLoader/es6-module-loader/jobs/48323472#L422) you'll notice the saucelabs auth always fails. 
It's because my code expect the saucelab username and access key to be exposed in the env (see https://github.com/ModuleLoader/es6-module-loader/pull/284#issuecomment-71205837).

So when I made my PR #284, I told Travis to [run the saucelabs tests out of the script command](https://github.com/ModuleLoader/es6-module-loader/pull/284/files#diff-354f30a63fb0907d4ad57269548329e3R19). So Travis will run them but their outputs wouldn't be taken into account (see [Travis Build Lifecycle](http://docs.travis-ci.com/user/build-configuration/#Build-Lifecycle)) like it will continue to fail as long as the authentication is not valid.

guybedford:
@douglasduteil ok right, thanks for explaining! So we can switch this back then?

douglasduteil:
> So we can switch this back then?

what "this" ?

douglasduteil:
The saucelabs test ?

guybedford:
Move these statements into script?

https://github.com/ModuleLoader/es6-module-loader/blob/master/.travis.yml#L20

douglasduteil:
You will need to expose the `SAUCE_USERNAME` and the `SAUCE_ACCESS_KEY` on the Travis environment first see [douglasduteil/dummy-fuzzy-batman#note-to-make-it-work](https://github.com/douglasduteil/dummy-fuzzy-batman#note-to-make-it-work)

guybedford:
Right, yes I've done that.

douglasduteil:
Alright then we're good :)
Move the lines and Ship It ! :shipit:

guybedford:
Great thanks, I've included them.

douglasduteil:
It's looking good :+1:

PR #307, opened 2015-02-12
Conceptual fix for #305

### LINKED ISSUE #305: asyncPartWayThrough should move proceed existing loads to next state.
First discussed here https://github.com/systemjs/systemjs/pull/347, there is a problem with `System.define` calls defining a load that has already been added to `loader.loads`.

@guybedford Any suggestions on the best way of creating a test for this?  The only idea I have is to:

1. Begin an import of "master" module that has a "slave" dependency.
2. Create a locate (perhaps fetch) hook to delay the "master" from resolving its "slave" dependency.
3. Within that hook, `System.define` the "slave" module.
4. Resolve the "master"'s locate (or fetch) hook.

Thoughts?

alexisabril:
Initial approach to fix #305 asyncPartWayThrough should move proceed existing loads to next state.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -361,7 +361,8 @@ function logloads(loads) {
         if (loader.loads[i].name == name) {
           existingLoad = loader.loads[i];
 
-          proceedToStepState(loader, existingLoad, stepState);
+          if(stepState.step == 'translate' && !existingLoad.source)
+            proceedToStepState(loader, existingLoad, stepState);
 
           return existingLoad.linkSets[0].done.then(function() {
             resolve(existingLoad);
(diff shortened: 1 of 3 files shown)


guybedford:
Nice I agree with the principle here.

Since this is a specific scenario, can we make the change rather to only check if the `state` is `translate`, the `load.source` is not already defined and then running the `proceedToTranslate` directly?

matthewp:
@guybedford I added 2769a9170b90a14c04d3941d7bd76ee08c7c1b6b to address your concern about only running this code to fix this specific case and also added a test which shows it works. Think the breaking test in travis is because I didn't commit the build artifacts.

guybedford:
Thanks, I've manually merged this in https://github.com/ModuleLoader/es6-module-loader/commit/67f45bfc2f1b3e7f64ef49cefe489ff87be1c580. Note that automated tests and manual tests are duplicated as we make the automated tests the primary ones. The file to edit is `system.spec.js`.

PR #312, opened 2015-02-20
need .some polyfill for IE8

tracer99:
The current module loader fails because loader.js calls load.dependencies.some on 276 which doesn't exist in IE8


guybedford:
The `some` code is only used within an assert / debugging statement, so I don't think it's worth the polyfill. We can add a conditional to the debugging statement not to run if Array.some isn't defined, which would be uglified out as well with the other assert statements fine - I usually manually define some but would definitely be nicer to have a working IE8 debugging experience, so thanks for this.

PR #315, opened 2015-02-22
Correct link to Babel in Readme

probins:
Existing link seems to just hang, so I've changed to the one linked to from babel's github repo.


guybedford:
Thanks.

PR #320, opened 2015-03-03
[1.0] Adjust the demo to the latest spec

nevir:
aka add `.js`. My understanding from the [module id -> address change](https://github.com/google/traceur-compiler/issues/1221) is that we should really be using relative paths here too; is that correct? (though both work)


addyosmani:
@guybedford It's my understanding based on a read through the [changes](https://github.com/google/traceur-compiler/commit/4dd20c56c9936880043ff366184059131c84640b) Traceur made to their modules repl that this is indeed correct. 

Merging away, but if @nevir and I have misread, please feel free to back out this change.

PR #325, opened 2015-03-07
Removed blocked JSX syntax

hipertracker:


guybedford:
See https://github.com/ModuleLoader/es6-module-loader/commit/4d0e42cbbf62274983f7764f3935060b6077048a.

Sorry for the frustration here. Got hung up on best-practises, but code should always allow workarounds.

PR #337, opened 2015-03-25
move sourceURL part to loader

mo22:
so that it also applies to already transpiled modules

I'm using es6-module-loader to load modules that are already transpiled - but still need the correct #sourceURL mapping for the eval.

### DIFF
--- a/src/polyfill-wrapper-end.js
+++ b/src/polyfill-wrapper-end.js
@@ -3,7 +3,11 @@
 // file to avoid adding those references to the evaluation scope.
 function __eval(__source, __global, __load) {
   try {
-    eval('(function() { var __moduleName = "' + (__load.name || '').replace('"', '\"') + '"; ' + __source + ' \n }).call(__global);');
+    // pass sourceURL tag so it will show up correctly in stacktraces
+    eval(
+      '(function() { var __moduleName = "' + (__load.name || '').replace('"', '\"') + '"; ' + __source + ' \n }).call(__global);' +
+      '\n//# sourceURL=' + __load.address + '\n'
+    );
   }
   catch(e) {
     if (e.name == 'SyntaxError' || e.name == 'TypeError')
--- a/src/transpiler.js
+++ b/src/transpiler.js
@@ -61,10 +61,6 @@
     var compiler = new traceur.Compiler(options);
     var source = doTraceurCompile(load.source, compiler, options.filename);
 
-    // add "!eval" to end of Traceur sourceURL
-    // I believe this does something?
-    source += '!eval';
-
     return source;
   }
   function doTraceurCompile(source, compiler, filename) {
@@ -89,10 +85,7 @@
       options.blacklist = ['react'];
 
     var source = babel.transform(load.source, options).code;
-
-    // add "!eval" to end of Babel sourceURL
-    // I believe this does something?
-    return source + '\n//# sourceURL=' + load.address + '!eval';
+    return source;
   }
 
 


guybedford:
I think rather have your transpiler output this correctly into the transpiled files individually.

The issue is that adding a sourceURL can mess with existing sourceURL or sourceMappingURL comments.

mo22:
Hmm, the problem is that this screws up relative module paths - i.e. when
the final path is not known at transpile time :(



On 25 Mar 2015, at 17:03, Guy Bedford <notifications@github.com> wrote:

I think rather have your transpiler output this correctly into the
transpiled files individually.

The issue is that adding a sourceURL can mess with existing sourceURL or
sourceMappingURL comments.

—
Reply to this email directly or view it on GitHub
<https://github.com/ModuleLoader/es6-module-loader/pull/337#issuecomment-86095917>
.

guybedford:
Source URLs work fine as relative URLs.

mo22:
Doesn't look like it for chrome at least - and even if so, the base url
would be the one of the file triggering the eval(), which would be
es6-module-loader.js, so relative paths would always be based on the path
es6-module-loaded.js is in.



On 25 Mar 2015, at 17:08, Guy Bedford <notifications@github.com> wrote:

Source URLs work fine as relative URLs.

—
Reply to this email directly or view it on GitHub
<https://github.com/ModuleLoader/es6-module-loader/pull/337#issuecomment-86097548>
.

guybedford:
Right - in SystemJS we actually enable this through a renormalization of the sourceURL, see https://github.com/systemjs/systemjs/blob/master/lib/extension-register.js#L37.

The goal would be to port this behaviour into here actually with the register extension.

mo22:
Hm, at that point to convert the sourceMappingURL to an absolute URL, i.e. the path to the source mapping file.

That does not change scripts URL, i.e. if you have a console.log(‘hello’);, chrome will usually tell you the filename in the console where this happened.
Without sourceURL and sourceMappingURL it would print (eval) or (VMxxx:1) or something.
With sourceURL it would print the value passed as sourceURL.
With only sourceMappingURL, and sourceMapping enabled in chrome dev tools, and sourceMappingURL being an absolute url, it would use the sourceMap to discover the correct filename.
With sourceURL _and_ sourceMappingURL, sourceMappingURL can be relative to sourceURL and the filename would be the one from the sourceMap, or, if sourceMaps are disabled, from the sourceURL.

So you could skip making sourceMappingURL absolute if you had doEval in polyfill-wrapper-end pass the sourceURL in line 66 to eval() - eval(source + ‘\nsourceURL=‘ + address + ‘\n’);, and also for node.js by passing a parameter to vm.runInThisContext(source, {filename: address});

guybedford:
Are you sure about sourceMappingURL taking preference over sourceURL? In my previous tests, I seem to remember adding sourceURL when a sourceMappingURL was present could cause the sourceMappingURL value to stop being used?

Otherwise, certainly, your suggestion makes a lot of sense.

mo22:
It is pretty much two different things:

sourceURL is parsed by eval() to identify the filename - i.e. for v8: there is StackFrame::GetScriptNameOrSourceURL() - http://izs.me/v8-docs/classv8_1_1StackFrame.html <http://izs.me/v8-docs/classv8_1_1StackFrame.html>
That value actually shows up in a caught exception (v8 Error object uses the sourceURL tag)

sourceMappingURL is parsed by the dev tools / debugger to retrieve the source mapping file and to deduce the original filename/line number for minified/transpiled sources.

for live environments, where you usually do not have a .js.map file hosted the dev tools would still have the sourceURL and line number (of the transpiled source file), that you could convert in the source file name/line number using the sourceMap.

guybedford:
Ok I'll make a note to test this change against your suggested adjustments to SystemJS in https://github.com/ModuleLoader/es6-module-loader/pull/337#issuecomment-86104789 and ensuring:

* Relative sourceMappingURL is normalized relative to sourceURL
* sourceURL doesn't interfere with sourceMappingURL source maps

If the above are ok then this will be great to merge.

mo22:
awesome, thank you!

mo22:
Just noticed that you are already using that at some points in system.js:

  __eval = function(source, address, sourceMap) {
    source += '\n//# sourceURL=' + address + (sourceMap ? '\n//# sourceMappingURL=' + sourceMap : '');

guybedford:
I just tested this out and unfortunately it breaks the line-by-line ES6 debugging in the test cases for me.

As you can tell, it's kind of a "what works" solution currently. Would love to be able to get this side of things on solid ground, so thanks for the input.

mo22:
Hmm, very strange. I’ll see if I can put some time into this later and make sure the tests work fine.

PR #341, opened 2015-03-31
Minimal Loader Build

guybedford:
This creates a minimal loader build of 4.6KB minified and gzipped.

Next step is to add System.register extension to that to give us the minimal production loader necessary.


src/register.js
```diff
@@ -0,0 +1,566 @@
...
+    return newDeps;
+  }
+
+  /*
+   * There are two variations of System.register:
+   * 1. System.register for ES6 conversion (2-3 params) - System.register([name, ]deps, declare)
+   *    see https://github.com/ModuleLoader/es6-module-loader/wiki/System.register-Explained
+   *
+   * 2. System.register for dynamic modules (3-4 params) - System.register([name, ]deps, executingRequire, execute)
+   * the true or false statement
```
probins: is something missing here? This doesn't make sense to me.

guybedford:
This has now been merged.

PR #346, opened 2015-04-04
0.17 wip

guybedford:
* New build style with minimal production build
* data-init removed
* ~~Promise no longer bundled~~
* Correct file:/// url form used in Node (three /)
* js extensions no longer added automatically
* baseURL can be a relative URL
* System.import takes parentName as second argument instead of an options object
* New spec normalization algorithm - all names normalized as URLs by default in normalize, with paths applied first if a plain module name.


douglasduteil:
Wa Sounds good !  :+1:

douglasduteil:
Just to konw, what led you to remove this built-in Promise polyfilling by default ? I was dreaming about it later ;)

guybedford:
Promises weren't implemented in browsers initially when this was implemented. Also when promises were implemented natively they were initially slower than the library implementations. I just tested yesterday and Promises are definitely now marginally faster than the when.js implementation, making it suitable.

guybedford:
Nearly ready to land here...

probins:
@guybedford What's the plan for future development? Are you planning further work for 0.18 or whatever, or are you going to be concentrating on 1.0 and the new spec?

guybedford:
@probins I've created a list of the blocking spec issues at https://github.com/ModuleLoader/es6-module-loader/issues/381. As soon as these are resolved, we'll either post a 0.18 or 1.0 on the new spec. 1.0 would be nice, but it really just depends on how close the spec is to stable.

probins:
ok, thanks. It's frustrating that nothing's happened on the spec for several months now.

PR #348, opened 2015-04-04
Track whether transpiler has ran with a flag on the Loader

### LINKED ISSUE #347: Transpiler breaks when traceur is already loaded and using multiple loaders
Because of the use of [this variable](https://github.com/ModuleLoader/es6-module-loader/blob/master/src/transpiler.js#L11) within the closure.  The fix would be to use something like `this.transpilerHasRan` as a flag instead, set on the Loader instance. I'll submit a fix.

matthewp:
This fixes a bug where setting Traceur and/or Babel wasn't happening in
multiple-loader scenarios. Instead of using a closure variable set one
on the loader itself.

Picked a long name to avoid collision. Fixes #347

### DIFF
--- a/src/transpiler.js
+++ b/src/transpiler.js
@@ -7,8 +7,6 @@
   function getTranspilerModule(loader, globalName) {
     return loader.newModule({ 'default': g[globalName], __useDefault: true });
   }
-  // NB this does not support sub-classing well
-  var firstRun = true;
 
   // use Traceur by default
   Loader.prototype.transpiler = 'traceur';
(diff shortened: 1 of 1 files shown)


src/transpiler.js
```diff
@@ -17,12 +15,12 @@
     var self = this;
 
     // pick up Transpiler modules from existing globals on first run if set
-    if (firstRun) {
+    if (!self.transpilerHasRan) {
```
guybedford: Spelling - `transpilerHasRun`.

guybedford:
Sure that works great for me.

matthewp:
Ok, updated.

guybedford:
Would you like a release now as well?

matthewp:
It's not needed, thanks though.

PR #350, opened 2015-04-07
Improve path matching

crisptrutski:
There were some issues with the existing code:

1. Contrary to documentation, "last match" would always win for wildcard matches
2. Exact matches worried about match length, which made no sense as this could match once only

Additionally, @guybedford suggested looking at "length" in terms of path depth (ie. number of slashes) rather than string length of the pattern.

This PR updates the `locate` function to address the 3 points above.

Another potential change mentioned by @guybedford would be "first match wins" for ties, although "last match wins" seems the right choice to me. Will add this if wanted though. In either case [the wiki](https://github.com/ModuleLoader/es6-module-loader/wiki/Configuring-the-Loader#paths-implementation) should be updated to mention this decision.

Left noise in separate commits, but happy to squash if suitable for inclusion.


crisptrutski:
D'oh wrong target

PR #362, opened 2015-04-28
newModule: only use getOwnPropertyNames if obj != null

matthewp:
If you do:

`System.newModule(null)` this code would have thrown without this fix.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -1020,7 +1020,7 @@ function logloads(loads) {
       var m = new Module();
 
       var pNames;
-      if (Object.getOwnPropertyNames) {
+      if (Object.getOwnPropertyNames && obj != null) {
         pNames = Object.getOwnPropertyNames(obj);
       }
       else {


guybedford:
Strictly `newModule` is designed to take a plain object as input. What is the use case for this?

matthewp:
We use it to prevent a module from being fetched in production (like css).  Did the old es6 spec say that it had to be a plain object?

guybedford:
I think that is what the top check is for (which was in the spec, although my typeof interpretation could have been wrong). I always write `newModule({})` for these cases myself, because it locks down the module, but other users may also be relying on this behaviour as well.

This is all changing, so I'm happy to add it if it helps.

matthewp:
Yeah, if you don't mind. I'll make a note to do it a different way in the future.  What is the whatwg/loader equivalent of newModule, Reflect.Module?

guybedford:
Still being specified (see wip at https://gist.github.com/dherman/fbf3077a2781df74b6d8) but most likely `Reflect.Module` yes.

PR #366, opened 2015-04-30
Add support to use TypeScript as a transpiler

vladima:
In this PR TypeScript dependency is added to the package.json as a GitHub url since version of TypeScript that supports System.register was not officially released yet. I'll send another PR once it will be available to public as npm package


dist/es6-module-loader.src.js
```diff
@@ -2493,6 +2508,16 @@ function logloads(loads) {
     return source + '\n//# sourceURL=' + load.address + '!eval';
   }
 
+  function typescriptTranspile(load, ts) {
+    var options = this.typescriptOptions || {};
+    options.module =ts.ModuleKind.System;
+    options.target = ts.ScriptTarget.ES5;
```
guybedford: In due course this should probably only be set if not already.

guybedford:
Nice! Happy to roll it out down the chain when this lands (branches here, SystemJS branches and jspm).

Just to discuss the dependency thing again - I'm still wondering if we should restrict System.register as a format to assume unique dependencies, to maintain the conceptual simplicity of a single setter running for a single dependency.

On the other hand it does also make sense to maintain the original dependency order, but as long as we still have the first mention order right, that loss shouldn't be an issue.

Perhaps you're right though and the correctness of dependency order is more important than a single-setter principle? Don't think it matters too much either way? Let me know what you think here - it's not really about the implementation itself, but what we should define for the format.

Note that Travis will run a build, so there is no need to include the "dist" changes in the commit.

I will do some testing as soon as I can in the new few days, but this is looking great to go at first glance.

pkozlowski-opensource:
@guybedford @vladima totally excited to see this work! Thank you!

MajorBreakfast:
It's nice to see typescript support. I know some people who're quite fond of it.
The current approach seems to replace traceur/babel with typescript. Since typescript is more than a transpiler, it's another language with it's own file extension `.ts`, it makes sense to consider allowing usage of babel/traceur side-by-side with typescript in the future.

pkozlowski-opensource:
Yeh, somehow this all in-browser transpilation should be pluggable so people can compose transformations (or drop them altogether if doing tranpilation as part of the build). This is what is happening in the 0.1 branch, no?

vladima:
@guybedford I would prefer to be as close to ES6 semantic as possible even at cost of increasing code size.

matthewp:
@pkozlowski-opensource System.translate is what you want.

guybedford:
@vladima sure I'm happy to get on board with the approach to the duplicate names.

Just to explain the delay here - I'm thinking we should run this against the 0.17 branch rather as I'm trying to get that release out as soon as possible and it is my top priority at the moment. Happy to manually do the merge as I do this as well, but just to explain why this hasn't landed yet.

vladima:
@guybedford sure, there is no rush, especially since official release of TypeScript 1.5 (where we've added support for System.register) has not been done yet. I would hold off merging this PR until the moment of the official release to have all npm dependencies in-place

guybedford:
@vladima it would be good to know when you think the release will be about to land - I'd like to try and get a consistent name for the `System.register` format if across TypeScript, Babel and Traceur. I'm not too keen on `system` because ultimately the format should be tied to the loader - `Reflect.loader.register(...)`, and not be SystemJS-specific (the `system` global is not defined in the spec anymore). I'm currently thinking of using `register`, although other ideas are very welcome.

guybedford:
@vladima to update, I'm happy to stick with the `system` name for now. For reference - https://github.com/google/traceur-compiler/pull/1916.

frankwallis:
I have been trying this out, and it is working really well. It's great that typescript is going to be a first class citizen in systemjs. I have found a small issue with the source maps, and so I have added some comments to the commits. thanks @vladima!

vladima:
@guybedford `system` sounds like a good name to me
Regarding source maps, I'm not sure that they are currently work correctly. AFAIR `//# sourceMappingURL` should be the last line of the file. This is not true for the current implementation
- `//# sourceURL=` is appended as the last line after the `transpile` step, not sure if debugger are resilient enough to allow these lines to be flipped
- `__eval` wraps transpiled code in function so  `//# sourceMappingURL` is definitely not the last line.

I've checked the ES6 module loader with Traceur as transpiler in Chrome and source maps didn't work.

frankwallis:
I think that SystemJS is taking care of that: https://github.com/systemjs/systemjs/blob/master/lib/extension-register.js#L34

I am running in SystemJS and they are working fine for me once the filename has been set.

vladima:
This means that probably I'm doing something wrong and I'll be glad to know what is it.
Let's say I have file 'app.js'

```javascript
import { Greeter } from 'greeter'

export function main(el) {
    let greeter = new Greeter(el);
    greeter.start();
}
```
and in my 'index.html' I'm doing 
```html
        System.transpiler = 'traceur';
        System.paths = {
            '*': '*.js',
            'traceur': '/node_modules/traceur/bin/traceur.js'
        }
        System.import('app').then(function(m) {
            var element = document.getElementById("content");
            m.main(element); // 1
        });
```

If I put a breakpoint on the line (1) and try to step into the function I can see that I'm stepping into the transpiled content, not in the original file.

IAlso in the debugger I can see that detected format for this file is 'es6' and it seems that function in SystemJS that was (mentioned before)[https://github.com/systemjs/systemjs/blob/master/lib/extension-register.js#L34] is not called for such modules (not sure if this is a bug or intended behavior)

frankwallis:
hmm, yes there may be an issue with the entry point. If you put a breakpoint in 'greeter' does that work?

vladima:
no, it does not and breakpoing in `register` function is not hit either.

frankwallis:
I agree there is something not quite right, but I'm having trouble reproducing the issue. What I see on my main app is that toggling the 'sourcesContent' option affects whether the source appears during the application bootstrapping phase (it needs to be true to see the source). Once the app is bootstrapped then everything is working fine. It is complicated somewhat because once the source has been loaded once successfully then the browser seems to serve up a cached version regardless.

I have created a cut down example [here](https://github.com/frankwallis/source-map-test) but in that one I can't make it go wrong, everything is working perfectly. Maybe you can try that and see if it works for you?

I also see what you mean about that code in SystemJS not getting run, but I don't know why that is happening either at the moment.

vladima:
@frankwallis  just tried to run your example on another box, seems to be working fine. I've updated the PR. 
Thanks!

vladima:
@guybedford I've talked with @mhegazy, it is ok not to wait  until the official release of TypeScript 1.5 and release ES6 module loader and related projects (SystemJS, JSPM) using Typescript dependency via github repro. After TS 1.5 will be officially released I'll send separate pull requests to ES6 module loader/SystemJS/JSPM to update dependency to npm,

vladima:
@guybedford is there anything else that you think should be addressed before merging this PR?

QuantumInformation:
Nice work!

vladima:
@guybedford if you want this PR to land in 0.17, should I close this PR and create a new one against 0.17 branch?

guybedford:
@vladima yes exactly that would be really great, but as I mentioned I'm happy to do this as well myself. This release is a complex one as it's a non-breaking rearchitecture as part of shifting towards a breaking change, but it's getting very close to ready.

vladima:
sure, I'll see what I can do on monday. Meanwhile I've checked out 0.17 and tried to run tests. It is expected that `npm run test:node` ends up with a bunch of errors like 
```
done() invoked with non-Error: Only file URLs of the form file:/// allowed
 running in Node.
```
?

guybedford:
You'll need to run `grunt` first on the module loader. The SystemJS tests are not yet passing in Node though.

vladima:
but do I understand correctly that test suite for es6 module loader itself should should be all green?

guybedford:
Yes it should be.

vladima:
@guybedford on my windows machine I get a bunch of test failures for both master and 0.17 branches. The error is the same for all failures:
```
 32) Custom Loader #import special #locate path rule should support special loading rules:
     Error: done() invoked with non-Error: Only file URLs of the form file:/// allowed running in Node.
```

The error is raised at [this line](https://github.com/ModuleLoader/es6-module-loader/blob/0.17/src/system-fetch.js#L54) when getting this url as an argument 'c://_git/es6-module-loader/node_modules/traceur/bin/traceur.js'. This url in turn was created [here](https://github.com/ModuleLoader/es6-module-loader/blob/0.17/src/system-resolve.js#L16) with 'traceur' as initial value for the name.

guybedford:
@vladima thanks so much for letting me know - yes that was a Node bug in Windows - I've fixed that up.

vladima:
great to see this fixed, thanks!

guybedford:
@vladima we've manually merged this into master now preparing for the 0.17 release.

I'm getting a strange error from TypeScript itself I was wondering if you could help me with?

```
 Error: Debug Failure. False expression: Output generation failed
    at Object.assert (/Users/guybedford/Projects/es6-module-loader/node_modules/typescript/bin/typescript.js:1379:23)
    at Object.transpile (/Users/guybedford/Projects
  es6-module-loader/node_modules/typescript/bin/typescript.js:36583:18)
      at SystemLoader.typescriptTranspile (/Users/guybedford/Projects/es6-module-loader/dist/es6-module-loader-dev.src.js:1253:21)
      at /Users/guybedford/Projects/es6-module-loader/dist/es6-module-loader-dev.src.js:1178:31
```

Let me know if you're able to assist with this at all.

vladima:
Yes, I'm aware of this issue, it is fixed in Typescript master. Today I'll update github repo with dependencies and check if this error go away (pretty sure it will)

Sent from Outlook<http://taps.io/outlookmobile>




On Tue, May 19, 2015 at 5:24 AM -0700, "Guy Bedford" <notifications@github.com<mailto:notifications@github.com>> wrote:


@vladima<https://github.com/vladima> we've manually merged this into master now preparing for the 0.17 release.

I'm getting a strange error from TypeScript itself I was wondering if you could help me with?

 Error: Debug Failure. False expression: Output generation failed
    at Object.assert (/Users/guybedford/Projects/es6-module-loader/node_modules/typescript/bin/typescript.js:1379:23)
    at Object.transpile (/Users/guybedford/Projects
  es6-module-loader/node_modules/typescript/bin/typescript.js:36583:18)
      at SystemLoader.typescriptTranspile (/Users/guybedford/Projects/es6-module-loader/dist/es6-module-loader-dev.src.js:1253:21)
      at /Users/guybedford/Projects/es6-module-loader/dist/es6-module-loader-dev.src.js:1178:31


Let me know if you're able to assist with this at all.

-

guybedford:
@vladima thanks that would be great!

vladima:
@guybedford I've pushed updated version of TypeScript package into the github repo, though tag stays the same (v1.5-beta2) so it will be great if you delete existing TypeScript package from node_modules and `npm install` to download updated version. However even with updated version I still get a number of failures in tests, all with the same error: 

``` 
TypeError: Cannot read property 'length' of undefined
    at Loader.CustomLoader.instantiate (C:\Sources\_git\es6-module-loader\test\custom-loader.js:113:29)
```

In [setupTranspilers](https://github.com/ModuleLoader/es6-module-loader/blob/ts-0.17/src/transpiler.js#L5) code creates a new module named 'typescript' that has uses `ts` from globals as  `default`. However [here](https://github.com/ModuleLoader/es6-module-loader/blob/ts-0.17/src/transpiler.js#L65) it is assumed that `__globals` will have property `typescript` which is not true - name in `__globals` will be `ts`. In general I'm not sure that I understand why do we go through the whole pipeline to load the transpiler when in theory we've already might be done all the work in `setupTranspilers` and transpiler module is already created and ready to use

guybedford:
@vladima thanks so much - the tests are all now passing fine for me on master!

About the global stuff, I added a change to allow the transpiler to always take preference as a global. This is not ideal though as the nicer modularity comes when we load transpilers always as modules. The issue here seems to be that TypeScript doesn't easily load as a global script in Node (behaving like it does in the browser). I guess we should probably install TypeScript as a npm package with the jspm Node conversion process though here? Would be good to ensure the override for jspm can handle both client and server loading to ensure consistency.

guybedford:
Merged and ready to go...

vladima:
@guybedford that is awesome, thanks! Can you please tell if these changes will be the part of 0.17 and when 0.17 is planned to be released (if this information is not confidential)

guybedford:
@vladima yes this is ready to go out with 0.17 now. Deadlines are fickle things, but the plan is to get the release out within this next week.

PR #377, opened 2015-05-18
TypeScript support (squashed)

crisptrutski:
Closes systemjs/issues#454

This is just #366 rebased on master. Have squashed things for convenience, but evidently messed up something - the transpiler is not loading correctly.

NOTE: @guybedford abusing closes here so that waffle groups things correctly, refs doesn't work, so watch out with merging


index.js
```diff
@@ -17,6 +17,10 @@ try {
   System.paths.babel = System.paths.babel || filePrefix + require.resolve('babel/browser.js');
 }
 catch(e) {}
+try {
+  System.paths.typescript = 'file:' + require.resolve('typescript/bin/typescript.js');
```
guybedford: This should be `filePrefix`

guybedford:
Awesome, thanks for this @crisptrutski. Finishing off the merge manually.

PR #382, opened 2015-05-28
Fix console-patching for IE8/9

steinerj:
Hi,

using version 0.16.6. we're getting a 'console undefined' error in IE9.
I've traced it back to this attempt to monkey-patch the console.assert function, which is... well, not working at all ;)

```javascript
if (__global.console)
    console.assert = console.assert || function() {};
```

so I replaced it with this:

```javascript
  if (!__global.console) {
      __global.console = { assert: function() {} };
  }
```

Hope, that's all right.

### DIFF
--- a/src/wrapper-start.js
+++ b/src/wrapper-start.js
@@ -4,9 +4,10 @@
   var isBrowser = typeof window != 'undefined' && typeof document != 'undefined';
   var isWindows = typeof process != 'undefined' && !!process.platform.match(/^win/);
 
-  if (__global.console)
-    console.assert = console.assert || function() {};
-
+  if (!__global.console) {
+      __global.console = { assert: function() {} };
+  }
+  
   // IE8 support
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks for this. Yes I gave up trying to polyfill assert a while back and have been manually patching it in IE myself.

guybedford:
...or just having devtools open

guybedford:
Note that these console statements are stripped in the minified source.

steinerj:
Thanks for merging! Not sure about the minified stripping, at least for 0.16.6. We initially got this error in production, using a minified version.

guybedford:
It has to be minified with the stripping option - there's definitely no `console.assert` in https://github.com/ModuleLoader/es6-module-loader/blob/v0.16.6/dist/es6-module-loader.js.

steinerj:
Ah, sorry my fault! We botched it. No worries, thank you for the quick merge and advice.

PR #383, opened 2015-05-29
Supply !eval suffix to transpilers directly.

Schibum:
As mentioned here https://github.com/systemjs/systemjs/issues/450#issuecomment-106772331 this fixes source maps with dynamically loaded and transpiled scripts.
Removed eval!-suffixed sourceURL, as this does not seem necessary anymore. I only tested debugging in Chrome though.


guybedford:
Thanks have fixed this separately on master.

Issue #386, opened 2015-06-02
Re-export performance


guybedford:
Re-exports in the System.register module format call the setter functions for each export successively such that the setter functions get run all the way to the top of the dependency tree of exponential order in the depth of the re-exporting.

Aurelia currently has re-exporting depths of at least 4, which means that 10 deep exports becomes of the order of 10^4 setter runs, which is causing performance issues for them.

The proposal to fix this is to provide a bulk-export variation of the `_export` function in the System.register format that takes any object containing key value pairs (optionally an entire module object for export *) as arguments instead of `key, value`.

This way:

```javascript
export * from 'p';
```

Can be written as:

```javascript
System.register(['p'], function(_export) {
  return {
    setters: [function(m) {
      _export(m);
    }]
    execute: function() {}
  };
});
```

Then in the implementation, this single export function means this triggers a single bulk export event up the tree depth making the setter count O(1^d) instead of O(n^d) for the number of re-exports.

Normal rexport statements can also use the same logic:

```javascript
export {p, q, r as s} from 'p';
```

can be written:

```javascript
System.register(['p'], function(_export) {
  return {
    setters: [function(m) {
      _export({
        p: m.p,
        q: m.q,
        s: m.r
      });
    }]
    execute: function() {}
  };
});
```

I'll aim to create an implementation for this in Traceur after this coming major release and include it in a coming patch. I may have overlooked something having not yet tested this theory out completely, so will report back when I've got full confirmation that the above optimization is working and we can include it in the other transpilers.

/cc @EisenbergEffect @sebmck @vladima

EisenbergEffect:
Thanks Guy! I really appreciate everyone's cooperation. Though this affects Aurelia pretty significantly, I think it's understood that this has pretty broad affects. Probably most people haven't realized it yet.

vladima:
@guybedford sure, I can add it to TypeScript once approach is confirmed to be valid and implementation details are availabl

vladima:
@guybedford I know that it is probably to early to ask about it but can I give an approximate time (month?) of release for the patch that will include bulk export? This will help us prioritizing tasks on the TypeScript side

guybedford:
@vladima thanks for your help with getting these features going. It's difficult to give a release date - there are a lot of things to align here. With some luck, it should be out before Friday. Note that the named bundle support is far more important than this issue to complete the loop on the TypeScript support.

guybedford:
[...]

PR #387, opened 2015-06-02
Implement bulk export variation

guybedford:
Possible implementation of https://github.com/ModuleLoader/es6-module-loader/issues/386 pending format testing.


src/declarative.js
```diff
@@ -134,15 +134,24 @@
       //    By disaling this module write-protection we gain performance.
       //    It could be useful to allow an option to enable or disable this.
       module.locked = true;
-      moduleObj[name] = value;
+
+      // export({name: value})
+      if (typeof name == 'object') {
+        for (var p in name)
```
maxnordlund: Don't you need a `hasOwnProperty` check here?

guybedford: Only if we expect to run in environments with Object.prototype extended with enumerable properties.

maxnordlund: What about `export(new MyObject)`, that has enumerable properties on its prototype. Won't this then copy those over to the module? I'm not to sure it will only receive plain objects, is all.

(If you extend `Object.prototype` I feel you should be on your own)

guybedford: This API is defined to take instances of Object only from transpiler outputs and is independent of user code.

guybedford: Specifically - Object or Module.

guybedford:
This is now tested and working against Traceur at https://github.com/google/traceur-compiler/pull/1957.

/cc @RobEisenberg

guybedford:
Merged.

EisenbergEffect:
Thanks Guy! Hopefully we can get the other transpiler implementors to move quickly.

Issue #388, opened 2015-06-03
Byte Order Mark included in fulfilled string from SystemLoader.fetch when executed in node


jonoward:
When SystemLoader.fetch is executed on node, if the source file is UTF-8 and has a BOM, this is not stripped out by default (see https://github.com/joyent/node/issues/1918). This is causing an issue in systemjs-builder, when minification is enabled, and the plugin-text is used, as the BOM gets encoded, so you land up with strings like:

`"\ufeff<div>hello</div>"`

I'll submit a PR for this shortly

guybedford:
@jonoward thanks again for this.

PR #389, opened 2015-06-03
Byte Order Mark included in fulfilled string from SystemLoader.fetch when executed in node

jonoward:
Fix for #388


guybedford:
Thanks!

Issue #378, opened 2015-05-20
Conformity with ES6 final draft


bjorm:
(I am just a "user" trying to find the best way to use ES6 today in a long-running project)

How valid is this project with regard to the final ES6 draft spec [1] ?

I found that chapter 26 referenced in loader.js [2]  does not appear in the ES 6 final draft, as are any other references to a `Loader`. The same goes for anything like `System.register`.

Am I not reading the spec right? I'll be the first to admit that I do not understand an awful lot of it.

Thanks a lot for any insight.

[1] http://wiki.ecmascript.org/doku.php?id=harmony:specification_drafts#april_14_2015_rev_38_final_draft
[2] https://github.com/ModuleLoader/es6-module-loader/blob/master/src/loader.js#L604

guybedford:
@bjorm no globals referring to the module loader are in the ES6 specification. Instead the ES6 specification only covers the semantics and syntax of ES6 modules. How they actually get loaded in browsers needs to be defined by a separate specification, which is currently tracking at the whatwg loader spec at https://github.com/whatwg/loader.

`System.register` is not related to either of the above specifications and is its own specification for an ES5 module format.

It's a difficult to understand space certainly. Better communication of the ideas is very much needed.

bjorm:
Thanks, that helped a bit.

Still, the prominent references to an outdated draft spec are not reassuring for potential adopters and people not familiar with all those working groups and committees involved in the specification process. 

> no globals referring to the module loader are in the ES6 specification. Instead the ES6 specification only covers the semantics and syntax of ES6 modules.

I do not doubt that. But the crucial thing for me as an outsider was that they were part of the draft spec referenced in the README and source files, but not of the final draft.

In my opinion the situation for outsiders would vastly improve if
- The references to the old draft spec in the README were changed to point to the whatwg loader spec or be kept the old reference but add an explanatory note along the lines of what you said above
- Remove all references in the source to outdated specs. From what you said, those in loader.js should point to the whatwg loader spec. I have no idea about the others.

I would be willing to give this a shot if you think it makes sense.

guybedford:
@bjorm at the very top of the README we have:

[...]

PR #390, opened 2015-06-04
Readme: extra clarification on status of loader spec

probins:
An attempt to help those like #378 who aren't aware of the background.


guybedford:
Sure, thanks.

PR #395, opened 2015-06-09
Remove reference to baseURL from Readme

probins:
as baseURL is no longer referred to in the config docs.

I think this removal/deprecation of baseURL should also be mentioned in the 0.17 release notes.


probins:
PS If baseURL is being deprecated, I'd suggest also mentioning this in systemjs release notes

guybedford:
Thanks, just moved it to SystemJS where it's still needed for bundling.

PR #396, opened 2015-06-11
Updates to readme

probins:
Proposes:
* additional text on module tag
* adding example of using Promise.all for multiple module import (https://github.com/systemjs/systemjs/issues/498#issuecomment-110807515)
* shuffling a few other bits about for a more logical sequence (IMHO)

Question: why is https://github.com/systemjs/systemjs/blob/master/docs/es6-modules-overview.md in systemjs? This surely belongs here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ This project implements dynamic module loading through `System` exactly to the p
 * Can be used as a [tracing tool](docs/tracing-api.md) for static analysis of modules.
 * Supports IE8+, with IE9+ support for ES6 development without pre-compilation.
 * The minified production loader is under 5KB minified and gzipped, making it suitable for production use, provided that modules are [built into ES5 making them independent of Traceur](docs/production-workflows.md).
+* Supports declaring modules with `<script type="module">`, the precursor of the proposed [`<module>` tag](https://github.com/dherman/module-tag/).
 
 For an overview of build workflows, [see the production guide](docs/production-workflows.md).
 
@@ -74,6 +75,10 @@ and load the module dynamically in the browser
 
 The dynamic loader returns a `Module` object, which contains getters for the named exports (in this case, `q`).
 
+See the [demo folder](https://github.com/ModuleLoader/es6-module-loader/blob/master/demo/index.html) in this repo for a working example demonstrating module loading in the browser both with `System.import` and with the module-type script tag.
+
+Although `System.import()` does not support the import of multiple modules defined in an array, because `System.import()` returns a Promise, this can be achieved by creating an array of `System.import`s and using `Promise.all()`.
+
 #### Setting transpilation options
 
 If using Traceur, these can be set with:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,19 +13,12 @@ This project implements dynamic module loading through `System` exactly to the p
...
 
 For an example of a universal module loader based on this polyfill for loading AMD, CommonJS and globals, see [SystemJS](https://github.com/systemjs/systemjs).
 
-### Documentation
-
-* [Configuring the loader](docs/loader-config.md)
-* [Production workflows](docs/production-workflows.md)
-* [Circular References &amp; Bindings](docs/circular-references-bindings.md)
-* [Extending the loader through loader hooks](docs/loader-extensions.md)
-* [Tracing API](docs/tracing-api.md)
```
guybedford: These links do need to be at the top. Perhaps make the first one an anchor link within the README if that's what you're after?

README.md
```diff
@@ -74,6 +67,17 @@ and load the module dynamically in the browser
...
 
+See the [demo folder](https://github.com/ModuleLoader/es6-module-loader/blob/master/demo/index.html) in this repo for a working example demonstrating module loading in the browser both with `System.import` and with the module-type script tag.
+
+Although `System.import()` does not support the import of multiple modules defined in an array, because `System.import()` returns a Promise, this can be achieved by creating an array of `System.import`s and using `Promise.all()`. For example:
+```js
+var promises = [];
+['m1.js','m2.js'].forEach(function(m){
+  promises.push(System.import(m));
+});
+Promise.all(promises).then(function(modules){...});
```
guybedford: This is unnecessary. There's no need to instruct users about ES6.

guybedford:
Thanks, note that because this repo provides no production workflow, it is not suitable for end-users and is more appropriate for implementors at this point. So the documentation is moving to reflecting that as well somewhat, hence why the basic ES6 module info was removed.

README.md
probins: These links just repeat what's in the previous section (in a more concise form), and it seemed to me that people coming to this file would want to read the getting started section first, and only then maybe want to move on to more details. So put the documentation links at the end

README.md
probins: ok, I can remove the example if you like

guybedford: Thanks, just been trying to keep the main pages small recently.

README.md
guybedford: The links at the top might be missed, with users wondering where the docs are - nice to make it easy to find on the page.

probins:
if you're moving general usage docs to systemjs, it might be better to have the note on Promise.all in there (somewhere)

guybedford:
In general, it is advisable to have a single main entry point over multiple import promises, so it's not something I'd want to encourage too heavily.

guybedford:
Thanks again for the suggestions - it's great to have this documentation feedback.

probins:
I'm using it for optional loads: parse querystring into an array and load the requested components with Promise.all. I don't think it makes sense in a static situation.

Issue #399, opened 2015-06-16
Fetch error stacks


guybedford:
Fetch errors don't show stacks properly, and just show the ENOENT error in Node.

We need to ensure there is a full stack.

Issue #397, opened 2015-06-12
Improve stacktrace for failed dependency loading


guybedford:
Originally posed by @OliverJAsh in https://github.com/systemjs/systemjs/issues/512:

This is what the errors looks like when a dependency fails to load:

```
Potentially unhandled rejection [3] Internal Server Error: <dependency>
	Error loading <dependency> from <requester>
	Error loading <dependency> (WARNING: non-Error used)
```

Sometimes `dependency` is a nested dependency of a dependency of `requester`, e.g. `requester -> foo -> bar -> dependency`. Is it possible to show the immediate requester (`bar`)?

nyarly:
Looking at an issue in systemjs/builder, I found that the right way to address that would be here. I think load objects would need to be extended to include their (first?) immediate requester, and their un-normalized requested name (i.e. "requested as") in order to report these errors usefully.

I've started (i.e. 15 minutes to fork, clone, write a failing test) work on a PR for this.

guybedford:
I think this can be done via some reflective lookups in the error handler itself to run a dependency trace just for the error message.

guybedford:
Thanks @nyarly for your work on this.

guybedford:
Released in 0.17.4.

PR #404, opened 2015-06-23
More informative load errors

nyarly:
Additions:

ENOENT errors are rebuilt as Error()s, so there's a stack from within ML.

Errors during load report the immediate source of the error and the pre-normalization name that triggered the error.

Meant to address #399 and #397


guybedford:
Thanks for looking into this. In my testing, `err.message` was not showing up in NodeJS-thrown errors, only `err.stack` hence https://github.com/ModuleLoader/es6-module-loader/blob/master/src/wrapper-start.js#L45. Any ideas on the Travis test failure?

nyarly:
It... looks like on Travis, the node generated ENOENTs are being produced as proper errors, although I also notice that the names of files haven't been normalized to file:/// 

It's entirely possible I was too conservative in my filtering (https://github.com/ModuleLoader/es6-module-loader/pull/404/files#diff-1037cb44c9524a2a57ee4bd6dd086837R70) and the thing to do is look for err.message? My impression locally has been that Node ENOENTs come back with a stack that is *only* the message, which is why I was creating a new error when they bubble into module-loader.

The behavior on Travis seems to be different. I've got node 0.10.36 and npm 2.5.0, vs. Travis's 0.10.39 and 1.4.28. If it weren't JavaScript I'd be surprised that those versions would make a difference in the stack of a basic error but ¯\_(ツ)_/¯

Let me take a whirl at checking for lack-of-message in the FS error instead of an presence-of-errno.

nyarly:
Seems to fix it: the custom loader tests hadn't been updated, so they were comparing to the e.stack instead of e.message. I don't know why I didn't catch that before push - I did do a merge which clobbered changed files in dist/

src/loader.js
```diff
@@ -191,6 +191,10 @@ function logloads(loads) {
       }
 
       load = createLoad(name);
+      if(!load.requests)
+        load.requests = [];
+
+      load.requests.unshift({as: request, from: refererName});
```
guybedford: Can we not avoid adding more metadata to the load record?

src/system-fetch.js
```diff
@@ -66,9 +66,13 @@
       else
         url = url.substr(7);
       return fs.readFile(url, function(err, data) {
-        if (err)
-          return reject(err);
-        else {
+        if (err) {
+          if (err.errno == 34) {
```
guybedford: Is there documentation somewhere on ENOENT not being an error object?

src/wrapper-start.js
```diff
@@ -37,18 +37,11 @@
...
-      if (isBrowser) {
-        newErr.message = err.message + '\n\t' + msg;
-        newErr.stack = err.stack;
-      }
-      else {
-        // node errors only look correct with the stack modified
-        newErr.message = err.message;
-        newErr.stack = err.stack + '\n\t' + msg;
-      }
+      newErr.message = msg + '\n\t' + err.message;
```
guybedford: I'm sure about reversing this ordering - the whole point is that the error starts with the source, and goes outwards through the calling sites. Not the reverse.

src/wrapper-start.js
```diff
@@ -37,18 +37,11 @@
...
+      newErr.stack = err.stack;
```
guybedford: Is this stack assignment strictly necessary? For Node only?

src/system-fetch.js
guybedford: I'm trying to come up with a test case that explicitly needs this. Any ideas?

guybedford:
Also I wonder if it is actually useful to throw non-errors in order to hide the stack as well. The fact that the stack originates from SystemJS internals is completely un-useful. Controversial I know though.

src/loader.js
nyarly: I couldn't see how to do it. I looked into the suggestion to trace-after-fail, but the trace itself fails and the after-trace code (which would report the trace) doesn't run in that case. I couldn't figure out how to get a sensible report out of a partial (i.e. failed) trace, even if the after-run code were used after a fail.

Possibly we could add a build option to enable load request tracking? Since the error report checks for the presence of the requests property, it'd fall back to the old behavior in its absence (that error message might refer to the option, though.)

src/system-fetch.js
nyarly: I couldn't find anything specifically. Googling for "node enoent" and variants is a flood of "I don't understand this error" rather than something about the message/stack thing. I did find this:

http://neversaw.us/nodejs-api/errors.html#errors_system_errors

Which clarified that ENOENT &co are simply augmented Error objects. Also clarified: the error.stack getter returns a string when it's called. (Foolish I was expecting an array.)

Based on the Error constructor, I suspect that (some versions) of Node or NPM have a bug in their construction of system errors, and they sub in the message for the stack.

As to testing: I specifically added tests about the non-existent file to ensure that it had a proper message and stack. Testing for e.g. errno 34  IMO is testing the implementation.

src/wrapper-start.js
nyarly: My thought was that the *message* was clearer as increasing specificity from left to right. "There was and error" "during loading" "this file" "which you called this" "from this caller" "caused by a missing file"

I completely agree that the *stack* should be in that order.

That said, while I think the increasing specificity of the message has clear merit, I'm not attached to it. I can swap it back if preferred.

src/wrapper-start.js
nyarly: Otherwise the new error will report it's source as coming from withing addToError, which is misleading, even if it's strictly correct.

nyarly:
For a library like es6-module-loader, I'd argue strongly that having a error trace back to its actual source (even within a concatenated file) is *extremely* useful. Especially with dynamic typing, it's almost impossible to properly enforce the library's interface, and violations show up as errors.

Case in point: I was driven to trace this error reporting back from systemjs-builder because the error messages I got were unhelpful. Even a stacktrace into es6-module-loader would have saved me a couple of hours.

Client applications like systemjs can catch errors and present them better e.g. by raising their own errors or throwing strings or whatever.

src/loader.js
nyarly: Expanding on this: if there were a request tracking option, I'd think that systemjs-builder would set it by default, but systemjs wouldn't (or wouldn't in browser, for instance.) It's *very* useful during development, but I can see the concern about wasted memory/weedy values in production.

nyarly:
@guybedford just checking in on what direction to take with this?

nyarly:
@guybedford Any thoughts on how to proceed here?

src/wrapper-start.js
guybedford: Great point, thanks.

src/wrapper-start.js
guybedford: Yes I'd prefer to treat it like a stack.

guybedford:
@nyarly thanks for your patience - just got back from some time off.

Just thinking about the loading trace issue, the point is that in general we can't know who loaded module X, because a given module can be loaded by any number of parents. So which parent chain do we report? And how to make this work with circular references? Coming up with a well-defined solution to this is half the battle surely?

It would help if you could perhaps clarify how your implementation would deal with these cases?

Thanks so much for your efforts here, it's really appreciated.

src/wrapper-start.js
nyarly: Okay, I understand this to mean "swap it back", which I'll do

nyarly:
Sorry - I had a comment here and it got lost somehow.

Reviewing a reflecting on this: the answer is "this first one" - these errors aren't holistic (although that might be a further goal: to report *all* load errors), they're immediate. What's reported is the first module that cannot be loaded, the immediate parent that was loading it and the name that parent is trying to use. The resulting exception crashes the whole load process - which is what happens as is.

guybedford:
Right, so I'm more than happy with that, and I think we might have the metadata already available (indirectly) to be able to construct such a graph at debugging time. I don't want to add new graph meta as mentioned above, so if possible it would be great to use the existing private API to work this out. Just let me know if you need any further help with this, very much appreciate the efforts.

nyarly:
I'm more than happy to take that on, but I haven't been able to find how to do that with any private API. As mentioned above, if the load fails, running a trace will fail as well, and you don't get any new information. Is there another interface I'm not seeing that might provide this information?

guybedford:
It should be possible to iterate over the loads in the linkSet, looking for the first load that is the direct parent of the load that has errored. This would be done via checking `parentLoad.dependencies` contains a dependency entry with a value equal to `load.name`. Let me know if that sounds workable here?

nyarly:
I'll certainly look into it - meant to have a answer before responding, but there have been distractions.

nyarly:
That worked great - I hadn't been able to pick apart the linkSet and load objects until now. New code doesn't build the request objects until an error is reported.

I *think* that the number of requests for a failed load should always be at most 1, since as soon as it fails, the load process is interrupted to report the failure. All of the requests should be found by the current code, so if there's a concrete failure case with more than one request, extension should be easy.

src/system-fetch.js
guybedford: The error for file not found looks like this for me in Node - 
```
[Error: ENOENT, open '/Users/guybedford/Projects/es6-module-loader/asdf'
	Error loading file:///Users/guybedford/Projects/es6-module-loader/asdf]
```

Any ideas why it looks like an array here?

Also if I removed the _errno_ check here, or have another error like an access error, the error looks exactly the same and is no different - so what is this code doing then? Or is it only for certain versions of Node?

guybedford:
Excellent, this is looking good to go soon! Just let me know about the error number detection?

Do you think you could you remove the build files from the PR? Just trying to avoid commit churn.

nyarly:
Okay: pulled the errno check and tests pass without it, so :shrug:. Reverted the dist/ files to master.

src/loader.js
```diff
@@ -516,13 +516,32 @@ function logloads(loads) {
 
   // 15.2.5.2.4
   function linkSetFailed(linkSet, load, exc) {
+    function requestsForLoad(){
```
guybedford: Space after `()`

src/loader.js
```diff
@@ -516,13 +516,32 @@ function logloads(loads) {
...
+      var reqs = [];
+      linkSet.loads.forEach(function(aLoad){
```
guybedford: Space after `(aLoad)`

src/loader.js
```diff
@@ -516,13 +516,32 @@ function logloads(loads) {
...
-
-      if (load)
+      if (linkSet.loads[0].name != load.name) {
+        requests = requestsForLoad();
+
+        if (requests[0]) {
+          var req = requests[0];
+          exc = addToError(exc, 'Error loading ' + load.name + ' as "' + req.as + '" from ' + req.from);
+        } else
+          exc = addToError(exc, 'Error loading ' + load.name + ' from ' + linkSet.loads[0].name);
```
guybedford: This should have braces if the first group has braces.

src/wrapper-start.js
```diff
@@ -37,20 +37,13 @@
...
-        // node errors only look correct with the stack modified
-        newErr.message = err.message;
-        newErr.stack = err.stack + '\n\t' + msg;
-      }
+      newErr.message = err.message + '\n\t' + msg;
+      newErr.stack = err.stack;
     }
     else {
-      newErr = err + '\n\t' + msg;
+      newErr = msg + '\n\t' + err;
```
guybedford: I think this order still needs to be reversed again to match the same semantics?

test/system.spec.js
```diff
@@ -211,11 +211,36 @@ describe('System', function () {
           .then(supposedToFail)
           .catch(function (e) {
             expect(e)
-              .to.be.equal('dep error\n\tError evaluating ' + baseURL + 'test/loads/deperror.js');
+              .to.be.equal('Error evaluating ' + baseURL + 'test/loads/deperror.js\n\tdep error');
```
guybedford: As in previous comment, this should be reversed. We start with the error, moving out through the call site reporting.

guybedford:
Thanks, I'm still seeing some diffs for the dist files though? Perhaps it could be worth squashing at this point.

src/wrapper-start.js
```diff
@@ -37,20 +37,13 @@
...
     if (err instanceof Error) {
       var newErr = new Error(err.message, err.fileName, err.lineNumber);
-      if (isBrowser) {
-        newErr.message = err.message + '\n\t' + msg;
-        newErr.stack = err.stack;
-      }
-      else {
-        // node errors only look correct with the stack modified
-        newErr.message = err.message;
-        newErr.stack = err.stack + '\n\t' + msg;
```
guybedford: I think this might have been here for a good reason.

Currently if I throw errors in Node from this branch, they look like:

```
[Error: ENOENT, open '/Users/guybedford/Projects/es6-module-loader/asdf.js'
	Error loading file:///Users/guybedford/Projects/es6-module-loader/asdf.js]
```

Which looks like an array not a stack.

Any ideas?

guybedford:
Thanks again. Sorry this isn't just any code so do need to be very careful about the details.

nyarly:
I'm working through these changes. I understand the "not just any code" point - which is why I was surprised when I started that there's no linting step (which would save some of this runaround) and that brace-less ifs are tolerated.

nyarly:
I think I figured out the dist file issue (my own idiosyncratic git configs... which I think I've corrected for this case.) Do you still want this squashed to master?

guybedford:
Thanks, yes please do squash the commits. Did you try out the error reporting in NodeJS? Would be interested to hear your opinions on that.

nyarly:
I did the squash, and wound up rolling back some of the error inspection stuff while I was there. The result is a bunch of fails in test:node - I'm digging in now to see what the details are.

For reference, which versions of node are you testing against?

guybedford:
Sorry I'm talking rubbish - it was actually just the way I was testing it via the fact that:
```
> Promise.reject(new Error('asdf')).catch(console.error.bind(console))
```

Gives an array-looking wrapper around the error. Wrapping a `throw` in a `setTimeout` makes the error look fine again.

In short, ignore my comments re Node errors. I can confirm 100% that the tests are all passing for me.

I think that just leaves squashing for merge then!

guybedford:
Hoping to do a release here soon. Let me know if you're able to clear this up, otherwise happy to jump in and merge this through manually as well.

nyarly:
So, I've re-merged master into this branch, but I can't get a rebase to work without wrecking the code. Part of that is the dist files in the tree, and part is my own inexperience with rebasing. Rather than hold this up, can I turn the rebase over to you?

guybedford:
Sure, no problem, thanks for your work here - I've manually merged this in https://github.com/ModuleLoader/es6-module-loader/commit/3201bb42a8f588b5ce6ca654b81f7e3849da1aa4.

PR #410, opened 2015-07-08
Restrict scopes of __moduleName

toddwong:
```javascript
// file foo.js
import bar from "bar"
console.log(__moduleName);
```
Without this patch, the code above will output "bar" instead of "foo" if this is the first time "bar" was imported.

### DIFF
--- a/src/transpiler.js
+++ b/src/transpiler.js
@@ -24,7 +24,7 @@ var transpile = (function() {
         transpileFunction = babelTranspile;
 
       // note __moduleName will be part of the transformer meta in future when we have the spec for this
-      return 'var __moduleName = "' + load.name + '";' + transpileFunction.call(self, load, transpiler) + '\n//# sourceURL=' + load.address + '!transpiled';
+      return '(function(__moduleName){' + transpileFunction.call(self, load, transpiler) + '\n})("' + load.name + '");\n//# sourceURL=' + load.address + '!transpiled';
     });
   };
 


guybedford:
Nice catch, thanks. Let's just make it a function argument then instead of a `var`?

src/transpiler.js
```diff
@@ -24,7 +24,7 @@ var transpile = (function() {
         transpileFunction = babelTranspile;
 
       // note __moduleName will be part of the transformer meta in future when we have the spec for this
-      return 'var __moduleName = "' + load.name + '";' + transpileFunction.call(self, load, transpiler) + '\n//# sourceURL=' + load.address + '!transpiled';
+      return '(function(){var __moduleName = "' + load.name + '";' + transpileFunction.call(self, load, transpiler) + '\n})();//# sourceURL=' + load.address + '!transpiled';
```
guybedford: Newline after the `})();` so the source map comment is on a fresh line.

toddwong:
Thanks guybedford!

Yes, function argument is better! 
Move the newline to after the `})();` fails some tests (maybe because the result of `transpileFunction.call(self, load, transpiler)` ends with a comment line?), so it may be better to add anothing newline after the `})();`  instead?

I don't know the GitHub thing very well, should I attach another patch to this PR? or just submit another PR?

guybedford:
Ah yes, two newlines then perhaps?

You can just add a new commit to the branch and it will show up on this PR. You can even rebase and force push the same branch to keep it as a single commit.

toddwong:
Done! Easier than I thought.

guybedford:
Thanks!

PR #413, opened 2015-07-15
Clarifying initial import

peteruithoven:
I never understood why I couldn't just use ES6's import in external scripts right away, I hope this comment will clarifies this for others. 
https://github.com/systemjs/systemjs/issues/586


guybedford:
Thanks. It's not quite the right place for it, but I've aimed to better clarify the loading distinction when describing the module tag at https://github.com/ModuleLoader/es6-module-loader/commit/87fe61bf5fa3d680fd3d45d7d2ecb6c8a720e59b.

PR #415, opened 2015-07-27
Add module/x-module-loader-module as Accept header

Munter:
Following up on the discussion here: https://gitter.im/systemjs/systemjs?at=55af92db9402748911923244

Adding an explicit Accept header allows server side middleware to identify requests made specifically to load modules, and thus to hook into the loader chain on the server side.

This is used in https://github.com/Munter/express-systemjs-translate but could potentially open up for a range of other tools to help optimize the development workflow

### DIFF
--- a/src/system-fetch.js
+++ b/src/system-fetch.js
@@ -47,6 +47,8 @@
       };
       xhr.open("GET", url, true);
 
+      xhr.setRequestHeader('Accept', 'module/x-module-loader-module */*');
+
       if (doTimeout)
         setTimeout(function() {
           xhr.send();


guybedford:
Awesome thanks!

//cc @caridy we're prototyping the Accept header work as @Munter had a use case for a server that can respond differently to module requests. Would be really keen to see something along these lines in the spec.

guybedford:
@Munter just thinking a little further about the ramifications here - SystemJS will accept `System.register` modules while the module loader only provides a guarantee of accepting ES6 modules. Perhaps it is worth renaming this to `module/x-es-module` and then having a separate Accept header in SystemJS for `module/x-system-register`? If this means duplicating the fetch implementation in the SystemJS repo that's fine. Happy to discuss further as well, we should be careful with the pattern we set up here.

guybedford:
Created https://github.com/ModuleLoader/es6-module-loader/issues/416 to track further discussion.

Munter:
Sounds reasonable. From a server implementors perspective the exact syntax matters very little since only a naive string match is needed. I'll follow your recommendations here

guybedford:
Ok thanks, I think we should go with the suggestion above then. The module format is part of what is accepted.

guybedford:
@Munter apologies to keep changing here but I've just added https://github.com/ModuleLoader/es6-module-loader/commit/29f41f47d4298ee76bb117258dbf131635063a74. `application/x-es-module` so it fits in the standard content types (http://www.w3.org/Protocols/rfc1341/4_Content-Type.html).

Munter:
No problem :)

I was thinking, the current Accept header is `application/x-es-module */*`, but shouldn't it rather be `application/x-es-module application/javascript`? Or does the future of es6-module-loader also allow non-javascript assets a responses?

guybedford:
Yeah that might make sense, although the module loader as specified only accepts ES modules. It's extensions to the module loader like through SystemJS that can enable other response types.

So I just ran a test in Chrome and my `<script>` tags are doing an `Accept: */*` instead of an `Accept: application/javascript */*`. Am I the only one getting this?

Munter:
I think `*/*` is standard behavior sadly

PR #420, opened 2015-07-31
Add System.transpiledSources

dmail:
This pull request let access the dynamically transpiled module sources. The main goal is to support source map in nodejs error stack trace.

If you want to see how we could support source map in nodejs check my repo at https://github.com/dmail/system-node-sourcemap.

### DIFF
--- a/src/transpiler.js
+++ b/src/transpiler.js
@@ -5,6 +5,7 @@ var transpile = (function() {
 
   // use Traceur by default
   Loader.prototype.transpiler = 'traceur';
+  Loader.prototype.transpiledSources = {};
 
   function transpile(load) {
     var self = this;
@@ -24,7 +25,7 @@ var transpile = (function() {
         transpileFunction = babelTranspile;
 
       // note __moduleName will be part of the transformer meta in future when we have the spec for this
-      return '(function(__moduleName){' + transpileFunction.call(self, load, transpiler) + '\n})("' + load.name + '");\n//# sourceURL=' + load.address + '!transpiled';
+      return self.transpiledSources[load.address] = '(function(__moduleName){' + transpileFunction.call(self, load, transpiler) + '\n})("' + load.name + '");\n//# sourceURL=' + load.address + '!transpiled';
     });
   };
 
--- a/src/transpiler.js
+++ b/src/transpiler.js
@@ -5,7 +5,6 @@ var transpile = (function() {
 
   // use Traceur by default
   Loader.prototype.transpiler = 'traceur';
-  Loader.prototype.transpiledSources = {};
 
   function transpile(load) {
     var self = this;
@@ -25,7 +24,7 @@ var transpile = (function() {
         transpileFunction = babelTranspile;
 
       // note __moduleName will be part of the transformer meta in future when we have the spec for this
-      return self.transpiledSources[load.address] = '(function(__moduleName){' + transpileFunction.call(self, load, transpiler) + '\n})("' + load.name + '");\n//# sourceURL=' + load.address + '!transpiled';
+      return load.transpiledSource = '(function(__moduleName){' + transpileFunction.call(self, load, transpiler) + '\n})("' + load.name + '");\n//# sourceURL=' + load.address + '!transpiled';
     });
   };
 
--- a/src/transpiler.js
+++ b/src/transpiler.js
@@ -24,7 +24,7 @@ var transpile = (function() {
         transpileFunction = babelTranspile;
 
       // note __moduleName will be part of the transformer meta in future when we have the spec for this
-      return load.transpiledSource = '(function(__moduleName){' + transpileFunction.call(self, load, transpiler) + '\n})("' + load.name + '");\n//# sourceURL=' + load.address + '!transpiled';
+      return '(function(__moduleName){' + transpileFunction.call(self, load, transpiler) + '\n})("' + load.name + '");\n//# sourceURL=' + load.address + '!transpiled';
     });
   };
 
--- a/src/loader.js
+++ b/src/loader.js
@@ -266,6 +266,7 @@ function logloads(loads) {
               load.declare = declare;
               load.depsList = deps;
             }
+            load.transpiledSource = transpiled;
             // empty {} context is closest to undefined 'this' we can get
             __eval(transpiled, load.address, {});
             curSystem.register = curRegister;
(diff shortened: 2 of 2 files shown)


guybedford:
Nice work! Do you really need to run your own project at https://github.com/dmail/node-sourcemap or could you not merge the undefined `prepareStackTrace` work with https://github.com/evanw/node-source-map-support?

Also perhaps a better than API than a `readSource` callback would be the expectation that the `Error` object contains an `originalSource` string? If such a thing was supported, I'd happily output that from SystemJS itself as it could form a neater protocol for this sort of use case. Let me know your thoughts?

dmail:
I'm very glad you like it :)

Concerning my own project, I'm not especially attached to it, but it was easier than doing a pull request to node-source-map-support. I would even prefer to merge with them but I have to take the time to understand their more numerous use case.

I love your idea of an `originalSource` property on the error object, if it comes to SystemJS I would immediatly change my code to get source from there. For information, the `readSource` API is inspired by the behaviour of node-source-map-support.

To conclude, I think that adding `System.transpiledSource` is still relevant even if I don't see an other use case at the moment.

dmail:
Moreover node-source-map-support used Error.prepareStackTrace to catch all errors and I cannot polyfill this behaviour. It makes me unconfident to send a pull request.

guybedford:
I must admit I'd be far more inclined to accept a PR to this project that adds the full node source maps support you have than I am to accept this `System.transpiledSources` API.

Also do you have any idea when projects like this won't be needed to get source maps support in future?

guybedford:
Or even the `error.originalSource` type protocol would be preferable.

dmail:
> a PR to this project that adds the full node source maps support you have

You want an other PR including node source map support in your code base?

> do you have any idea when projects like this won't be needed to get source maps support in future?

I hope as soon as possible but I don't have information

> Or even the error.originalSource type protocol would be preferable.

I'm ok to rely on this, but we need to know the transpiled sources to populate `error.originalSource`. We could make `System.transpiledSources` private but we need something it, no?

dmail:
Imagine the following code : 

test.js
```javascript
export function test(){
   throw new Error();
}
```

index.js
```
var System = require('es6-module-loader');

System.import('./test.js', function(exports){
  exports.test();
}).catch(function(error){
   // your code cannot detect this error but you may still want to support sourcemap here
the developper have to manually wrap the error
   // you have to manually add it like so
   require('system-node-soucemap')(error);
   throw error;
});
```

The response

guybedford:
Ah right, I see what you mean.

Ok, well then I'd suggest instead of a protocol or including the code in the module loader, it may be worth having a source maps tool that injects itself into the module loader:

```javascript
function injectIntoLoader() {
  var transpiledSourcesStore = {};
  var systemTranslate = System.translate;
  System.translate = function(load) {
    return systemTranslate.call(this, load)
    .then(function(source) {
      transpiledSourcesStore[load.name] = source;
      return source;
    });
  }
}
```

dmail:
Adding a hook for System.translate was my first intent from the very beginning.
"Unfortunately", transpilation occurs after translate, during the instantiate phase, as you can see here : https://github.com/ModuleLoader/es6-module-loader/blob/master/src/loader.js#L260.

guybedford:
@dmail what about just setting `System.trace = true` and using the trace API then?

dmail:
@guybedford If each load.source is the transpiled source it seems like a nice way to do it.

I'm not sure of how to combine trace API, module instantiation & the call to the function `makeErrorSupportSourceMap`. Can you provide a short example to help me?

guybedford:
Yes `load.source` is the transpiled source in the trace. Then you just lookup the load in `System.loads` via the address just like you would have done in transpiledSources. See https://github.com/ModuleLoader/es6-module-loader/blob/master/docs/tracing-api.md for more info.

dmail:
Thank you, we are almost done here but `load.source` returns the untranspiledSource, see https://github.com/ModuleLoader/es6-module-loader/blob/master/src/loader.js#L597.
Could we  add `transpiledSource` property on each load record ?

src/transpiler.js
```diff
@@ -24,7 +24,7 @@ var transpile = (function() {
         transpileFunction = babelTranspile;
 
       // note __moduleName will be part of the transformer meta in future when we have the spec for this
-      return '(function(__moduleName){' + transpileFunction.call(self, load, transpiler) + '\n})("' + load.name + '");\n//# sourceURL=' + load.address + '!transpiled';
+      return load.transpiledSource = '(function(__moduleName){' + transpileFunction.call(self, load, transpiler) + '\n})("' + load.name + '");\n//# sourceURL=' + load.address + '!transpiled';
```
guybedford: Can we move this out of `transpiler.js` and to the call site that calls `transpile`?

guybedford:
I'd consider adding it. Note though that the internal transpilation of the module loader here is not a production workflow and is only for experimentation. For example, SystemJS would support source maps in Node work against `load.source` exactly as you have currently without this change.

dmail:
> Note though that the internal transpilation of the module loader here is not a production workflow and is only for experimentation.
Under which conditions you would remove transpilation from module loading ?

> For example, SystemJS would support source maps in Node work against load.source exactly as you have currently without this change
How could they do it ?

My last commit moves transpiledSource from transpiler.js to loader.js. Thank you for you time and your amazing work @guybedford

guybedford:
SystemJS does the ES6 -> ES5 transpilation within the transpile hook. The transpilation layer in this project is only really for experimental dev workflows and not recommended for use in production. To support SystemJS you would need to make sure you support `load.transpiledSource || load.source` as it wouldn't set `load.transpiledSource` as this code wouldn't ever run.

I'd prefer to stick with the SystemJS route and not encourage source maps for this workflow. Let me know if that is a possibility for you at all?

dmail:
> SystemJS does the ES6 -> ES5 transpilation within the transpile hook

Indeed, as the following line shows : https://github.com/systemjs/systemjs/blob/a38da7ed4c1c66e853c20d83375d35100f3453f8/lib/esm.js#L31.

> I'd prefer to stick with the SystemJS route

I think like you, transpilation should occur during the translate hook.

>  Let me know if that is a possibility for you at all?

At first, I wanted to avoid systemjs because it's more high level than es6-module-loader but it's totally ok for me.

I've updated my module to rely on the translate hook (SystemJS). SystemJS can now have nice error stack on node, oh joy !

guybedford:
Awesome, thanks for understanding. Can I close this then?

dmail:
Let's close this, thank you very much for your time

PR #423, opened 2015-08-07
Allow for loading source maps for files loaded by XHR

bragradon:
Hi all,

For the application I work on we load source maps via the [SourceMap response header](https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit#).
This doesn't appear to work currently for files that are loaded via XHR.
From what I can tell XHR loaded source is eval'ed during the instantiation phase but because there is no sourceMappingURL comment in the file the browser doesn't know should be a source map.

I attempted to write a plugin to hook the fetch method:

```javascript
    var system_fetch = System.fetch;
    System.fetch = function(load) {
        var System = this;
        return new Promise(function(resolve, reject) {
            system_fetch.call(System, load).then(function(response) {
                var xhr =
                    ('metadata' in load) &&
                    ('fetch' in load.metadata) &&
                    load.metadata.fetch.xhr;

                if (xhr) {
                    var sm_header = xhr.getResponseHeader('SourceMap');
                    if (sm_header) {
                        response += ';//# sourceMappingURL=' + sm_header;
                    }
                }

                resolve(response);
            }, function(err) {
                reject(err);
            });
        });
    };
```

However, this doesn't work as I do not have access to the XHR object.
This pull request adds the XHR object into the load.metadata so that the above can work

### DIFF
--- a/src/system-fetch.js
+++ b/src/system-fetch.js
@@ -23,7 +23,10 @@
         doTimeout = true;
       }
       function load() {
-        fulfill(xhr.responseText);
+        fulfill({
+          source: xhr.responseText,
+          responseHeaders: xhr.getAllResponseHeaders && xhr.getAllResponseHeaders()
+        });
       }
       function error() {
         reject(new Error('XHR error' + (xhr.status ? ' (' + xhr.status + (xhr.statusText ? ' ' + xhr.statusText  : '') + ')' : '') + ' loading ' + url));
@@ -70,7 +73,7 @@
           if (dataString[0] === '\ufeff')
             dataString = dataString.substr(1);
 
-          fulfill(dataString);
+          fulfill({ source: dataString });
         }
       });
     };
@@ -82,5 +85,11 @@
   SystemLoader.prototype.fetch = function(load) {
     return new Promise(function(resolve, reject) {
       fetchTextFromURL(load.address, resolve, reject);
+    }).then(function(result) {
+      if (result.responseHeaders) {
+        load.metadata.responseHeaders = result.responseHeaders;
+      }
+
+      return result.source;
     });
   };


guybedford:
Thanks, sure I'd be happy to work with something along these lines. Perhaps let's rather make the `fetch` function return a `{ source, sourceMap }` pair, then set `load.metadata.sourceMap` if the sourceMap option is defined.

bragradon:
Just clarifying; currently the `fetchTextFromURL` function is responsible for just fetching the contents of the url it is provided. Your proposal seems to imply that it would additionally return the sourceMap of the content if it is available.

I am confused as to whether you want me to return the content of the sourceMap or just the url link to the map, else should the variable be called `sourceMapHeaderContent`? My worry is about causing confusion when others read the code and instead think that either the sourceMap itself or that the content of the `//# sourceMappingURL=` comment  is being returned.

For my use case, I only need the header content which I can then add to the bottom of the retrieved content in my hook. This is so that browsers can control the retrieval of the sourceMap, i.e. only when DevTools are open. I wouldn't want to perform unnecessary XHR calls (For my app, we don't even serve source maps in our production environment unless debugging is enabled on the server).

So, proposal:
- Within `fetchTextFromURL`, on success: `resolve({source, sourceMapHeaderContent})`
  - So we check the response header, look for SourceMap header and just return whatever it has
- Add .then to Promise within `SystemLoader.prototype.fetch` (approx line 95)
  - Check and populate `load.metadata.sourceMapHeaderContent`, if applicable

I think I need a better variable name.

guybedford:
Ok sure, that sounds great then. Perhaps `sourceMapURL`?

guybedford:
Maybe `sourceMappingURL` for consistency with the browser name.

bragradon:
Changes uploaded. 
1 thing to note: there is a potential that someone might think that load.metadata.sourceMappingURL might contain the contents of `//# sourceMappingURL=` comment. Is there anywhere to add explicit information for this metadata?

I think including the contents of the comment would be a bit burdensome; we'd need to scan all incoming sources for the comment and the have to handle cases where a source has the comment more than once (as well as false positives). I think this complexity is best avoided.

guybedford:
Thanks for handling the edge cases. It would also be worth checking that `getResponseHeader` doesn't throw an error when using CORS. Perhaps try testing against say jQuery from the jQuery CDN, and see if that works ok? It may well throw an access control error there that we need to be catching.

Yeah I do see your point, perhaps `sourceURLHeader` is the best option after all.

bragradon:
Checking the CORS behaviours with jQuery CDN:
What happens is the browser will attempt to get the header and, as the header is unsafe, will show as error in the console (Chrome displays: 'Refused to get unsafe header "SourceMap"') and the function returns null. This appears to be the behaviour in all browsers I checked (FF, IE10, Safari8, Ch).

guybedford:
Ahh, that is somewhat problematic - it may be worth reconsidering then. We could do a preflight etc, but then we don't know which servers would support it. We could also not check the header when we know it is a request to a different origin, but then we're significantly reducing the value of the use case to just local files. What are your thoughts?

bragradon:
We could alternatively return all headers via `xhr.getAllResponseHeaders()`.
This function will return a single string containing all the headers in the response.
For a cross-origin request it will just return the valid readable headers, thus no error in the console.
So the `fetchTextFromURL` function would instead return {source, responseHeaders}.

I could then process the headers string into its constituent parts and check for the SourceMap header.
or would you want responseHeaders to be an object, ala:

```javascript
function load() {
  var headers = xhr.getAllResponseHeaders && xhr.getAllResponseHeaders();
  if (headers) {
    var headers = headers.trim().split('\n').reduce(function(obj, header) { 
      header = header.split(':');
      var key = header.shift().trim();
      var val = header.join(':').trim();
      obj[key] = val;
      return obj;
    }, {});  
  }
  
  fulfill({
    source: xhr.responseText,
    responseHeaders: headers
  });
}
```

guybedford:
I do worry somewhat here about the cost of re-parsing all headers. Consider an app with 500 requests now having 500 headers parsed. I guess it's not that much of a cost, but we do try and do as little processing as possible, as this code has to have high performance. I'm tempted to say we should stick with only doing this for same-origin requests, or leave out the feature as something that requires a custom fetch extension to the loader.

bragradon:
Performance is a valid concern here especially since it would be for every request and not everyone is going to want or need the headers.
I'm quite happy to do the parsing as part of a custom fetch extension and just leave this as:

```javascript
function load() {
  fulfill({
    source: xhr.responseText,
    responseHeaders: xhr.getAllResponseHeaders && xhr.getAllResponseHeaders()
  });
}```

guybedford:
Thanks for understanding. So I was happy to accept a `sourceMap` option because that is enabling better debugging, which is fine as an exception to the spec in the implementation. But to allow all the headers to be added to metadata creates an API that won't survive against the real spec, so it's not something we should make available when it will be taken away.

So I'd actually suggest doing a custom build of the module loader with your own fetch implementation. See the SystemJS makefile for an example of how to do this. https://github.com/systemjs/systemjs/blob/master/Makefile#L53.

bragradon:
Ah ok, yes that is an alternate approach I can take.
I was attempting to get the change upstream so I wouldn't need to worry about potential future changes, as well as being able to use it within system.js. But if this will cause spec issues then you're right; providing a something just to take it away will only cause headaches.

It does make me worry a little about whether the spec will handle some of the use cases I may have.
Besides this one (SourceMap header), I've also been looking at using Content-Type so to perform additional processing based on response mime-type. It could be that I've misunderstood the spec and that in my case I'd be extending the fetch method regardless.

Would it still be acceptable to provide the SourceMap header only for same-origin requests?

guybedford:
@bragradon the spec is exactly designed to handle these sorts of custom cases by allowing a custom `fetch` hook in the first place. With `fetch` API having wider browser support in future, a custom fetch hook would be even less code.

I think I'd rather leave it out at this point actually, as it's not ideal to encourage features which rely on same-origin support, splitting workflows for asset server separation. Just let me know if you need any help with getting a custom build going.

PR #432, opened 2015-09-20
Add if else condition to prevent unspecified error on IE

davidecantoni:
This change prevents the following error in IE if loading an external resource like jQurey.
Tested on IE8 Win7 + IE9 Win7 and fixes the following problem:
```
SCRIPT16389: Unspecified error.
```

### DIFF
--- a/src/system-fetch.js
+++ b/src/system-fetch.js
@@ -50,12 +50,13 @@
         }
       }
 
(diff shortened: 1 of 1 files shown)


guybedford:
Excellent thanks! I was wondering what that was.

PR #435, opened 2015-09-30
Removing the sans-promises files, since they are not in the master br…

joeldenning:
…anch and are not generated by grunt. Let me know if there is some other reason to keep them around :)


guybedford:
:+1:

PR #437, opened 2015-09-30
Updating numbering throughout, and rearranging code accordingly

joeldenning:
So now the specification numbers are correct and the code is in the file in the same order as the specification. I put in some TODOs for some things that have changed in the spec, as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ bower_components
 tmp
 coverage
 .DS_Store
+npm-debug.log
--- a/src/loader.js
+++ b/src/loader.js
@@ -23,7 +23,6 @@
     };
   }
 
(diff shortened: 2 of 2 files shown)


guybedford:
Amazing thank you! Can you update the diff to remove the `dist` changes? The tests should still build fine.

guybedford:
:+1: for merge.

joeldenning:
Just removed the changes to the dist files. But doesn't that mean that the tests will not actually be testing the source code?? Do you usually only update the dist files when releasing?

guybedford:
Yes exactly, typically only builds are done for releases. The tests run the build first at https://github.com/ModuleLoader/es6-module-loader/blob/master/.travis.yml#L35.

PR #439, opened 2015-10-08
Started work on Registry constructor, including making instances iterable

joeldenning:
I plan on sending some more pr's that really integrate Registries and their functions (lookup, install, uninstall, etc) into the existing tests and codebase. But for now I've left those functions where they are (on Loader.prototype) and started small with just the Registry constructor.


joeldenning:
Looks like the tests are failing because I am using `for ... of` syntax in the tests. That has been turned on by default since Chrome 38 (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of), but unfortunately the version of chrome that ships with the Ubuntu that Travis uses is Chrome 37 (https://github.com/travis-ci/travis-ci/issues/3354#issuecomment-82070987).

Additionally, it looks like the versions of Firefox and Chrome that are being used might not be letting me set the Symbol.iterator of objects.

I'm not super familiar with how to get Travis to use newer browser versions -- do you know how much effort is involved? The reason for using Symbol.iterator is to meet the whatwg/loader spec, and the reason for using `for ... of` is to be able to test it.

src/loader.js
```diff
@@ -216,6 +216,38 @@
...
+        var registryEntryIndex = 0;
+        return {
+          next: function() {
+            if (registryEntryIndex < instance._registry.registryData.length - 1) {
+              return {
+                value: instance._registry.registryData[registryEntryIndex++],
+                done: false
+              }
+            } else {
+              return { done: true };
```
guybedford: To be pendantic should there not be a `value: undefined` here to preserve the signature?

src/loader.js
```diff
@@ -216,6 +216,38 @@
...
+  // For now, registry objects are a work in progress that don't fully integrate into the rest of the code base
+  function Registry(loader) {
+    if (!this.constructor)
+      throw new TypeError('The Registry constructor must be called with "new"');
+    if (typeof loader !== 'object')
+      throw new TypeError('The Registry constructor must be called with a loader object');
+    this._registry = {
+      registryData: [],
+      loader: loader
+    }
```
guybedford: semicolon

src/loader.js
```diff
@@ -216,6 +216,38 @@
...
+    //4.4.2
```
guybedford: spaces after comment syntax?

guybedford:
Perhaps just pull in core-js to polyfill the iterator functionality for the tests?

src/loader.js
```diff
@@ -216,6 +216,38 @@
...
+      var instance = this;
+      this[__global.Symbol.iterator] = function() {
+        var registryEntryIndex = 0;
+        return {
+          next: function() {
+            if (registryEntryIndex < instance._registry.registryData.length - 1) {
+              return {
+                value: instance._registry.registryData[registryEntryIndex++],
+                done: false
+              }
```
guybedford: semicolon

src/loader.js
```diff
@@ -216,6 +216,38 @@
...
+            } else {
```
guybedford: I generally do a newline here, but I understand it is personal preference.

src/loader.js
```diff
@@ -216,6 +216,38 @@
...
+              return { done: true };
+            }
+          }
+        };
+      }
```
guybedford: semicolon

src/loader.js
```diff
@@ -216,6 +216,38 @@
   };
 
   // 4. Registry Objects
+  
+  // 4.2.1
+  // For now, registry objects are a work in progress that don't fully integrate into the rest of the code base
+  function Registry(loader) {
+    if (!this.constructor)
+      throw new TypeError('The Registry constructor must be called with "new"');
```
guybedford: Do we really need this error?

guybedford:
It's really amazing to see these changes moving, thank you!

src/loader.js
joeldenning: MDN indicates that `value: undefined` is optional, but I put it in the next revision as you suggested, for clarity.

src/loader.js
joeldenning: done

src/loader.js
joeldenning: done

src/loader.js
joeldenning: done

src/loader.js
joeldenning: I don't mind either way -- changed to follow the rest of the repo (with a new line)

src/loader.js
joeldenning: done

src/loader.js
joeldenning: I put it in because it's part of the spec (4.2.1 #1), but am happy to take it out if you'd prefer.

joeldenning:
I brought in the core-js polyfill to the tests and it helped with Symbol.iterator, but there are still some problems:

-Chrome 37 still doesn't like the `for ... of` syntax in registry.spec.js (https://travis-ci.org/ModuleLoader/es6-module-loader/jobs/84425307#L952)
-The corejs polyfill appears to be overwriting Symbol.iterator so that it no longer is really @@iterator anymore? (https://travis-ci.org/ModuleLoader/es6-module-loader/jobs/84425307#L960)
-Isn't it risky to throw polyfills into the test environment that are not going to be there when real dependents of es6-module-loader may not have those polyfills?

Let me know if I misunderstood your suggestion with core-js.

src/loader.js
guybedford: Sorry missed that, correctness is best as long as its not too much at the cost of performance or size!

src/loader.js
guybedford: Thanks, maintaining the return value object signature of properties and types allows the function to be best optimized by hidden classes in the compiler, as far as I'm aware.

guybedford:
It would be good to have a flag in Karma (like we have a flag for IE8 in https://github.com/ModuleLoader/es6-module-loader/blob/1.0/karma.conf.js#L24 which is a separate test in Travis from https://github.com/ModuleLoader/es6-module-loader/blob/1.0/.travis.yml#L22), to use the native iteration protocol. Then by default we should use a polyfill environment to be sure that works as well. Unfortunately because we don't provide `.entries()` (it's not in the spec yet, but I believe it may well be), I guess there's no way to test in a polyfill environment easily? Or can we just test the desugar of a direct call of the iteration protocol?

guybedford:
(the custom browser launchers for IE8 is specified at https://github.com/ModuleLoader/es6-module-loader/blob/1.0/karma.conf.js#L118, something similar for just modern Chrome / etc should be possible here?)

joeldenning:
I put the native iterator test into its own file so it can be excluded/included in different test environments. However, I'm having trouble getting Travis/Sauce Labs to use newer versions of chrome, so I'll have to look into that when I get the chance

joeldenning:
I had to comment out the travis tests for --native-iterator because I couldn't figure out how to get travis/sauce labs to launch newer versions of chrome/firefox that support the native iterator.

For now, though, I left the --native-iterator tests in the karma.conf so that they can be run from the command line and perhaps resurrected in the travis tests at some point. Let me know if you'd prefer me to remove those tests altogether and I can do that.

karma.conf.js
```diff
@@ -123,6 +127,19 @@ module.exports = function(config) {
...
+    if (options['native-iterator']) {
+      customLaunchers = {
+        SL_Chrome: {
+          base: 'SauceLabs',
+          browserName: 'chrome'
+        },
+        SL_Firefox: {
+          base: 'SauceLabs',
+          browserName: 'firefox'
+        }
```
guybedford: Did you try setting the browser `version` property like in https://docs.saucelabs.com/reference/test-configuration/#version-browser-

guybedford: Actually I guess the docs to read here are actually https://github.com/douglasduteil/gesalakacula.

guybedford:
Thanks for following this through.

I'm happy to leave the test in to be resurrected later.

Let me know if the new doc link helps, otherwise will go ahead and merge.

karma.conf.js
joeldenning: I tried both doing it with gesalakacula and without gesalakacula. And when I tried it without, I started out using a specific "version" property but then removed the property because it didn't seem to work. See https://travis-ci.org/ModuleLoader/es6-module-loader/pull_requests for all of my attempts at getting this thing to pass :)

The reason why you can't see all of those commits is because I was squashing and force pushing, in an effort to keep this pr at a reasonable number of commits.

joeldenning:
Those links to the documentation are some of the ones I was looking at when I was trying out different things. There is probably something obvious that I missed as I was doing it, but I wasn't able to figure it out. I even reached out to SauceLabs and pointed them to my code to see if they could figure it out, but wasn't able to get an answer from them.

karma.conf.js
joeldenning: The latest revision that does not have a "version" property was an attempt to use the latest version of Chrome/Firefox, because when you leave out the version it's supposed to use the latest chrome. (as explained in the link that you pointed me to)

guybedford: I think you only need one launcher called `Evergreen` with `{ firefox: 'dev', chrome: 'dev'` }`?

joeldenning:
I tried using a few more configurations, based on the one you suggested:

```
customLaunchers = {
        Evergreen: {
          firefox: 'dev',
          chrome: 'dev'
        }
      }
```
https://travis-ci.org/ModuleLoader/es6-module-loader/builds/86239198

```
customLaunchers = {
        Evergreen: {
          base: 'SauceLabs',
          firefox: 'dev',
          chrome: 'dev'
        }
      }
```
https://travis-ci.org/ModuleLoader/es6-module-loader/builds/86318193

```
customLaunchers = geSaLaKaCuLa({
        Evergreen: {
          firefox: 'dev',
          chrome: 'dev'
        }
      })
```
https://travis-ci.org/ModuleLoader/es6-module-loader/builds/86318950

All of them failed, so I re-commented out the travis tests to get the build passing again.

guybedford:
@joeldenning sure I'll put some time aside to look into getting these tests running next week.

guybedford:
Finally managed to take a look at this and it was as simple as replacing `Evergreen` with `Windows 7` as it must be a valid platform name.

See https://travis-ci.org/ModuleLoader/es6-module-loader/builds/88637168 for the passing build.

Let me know if you want to do anything else here, otherwise I will go ahead and merge this with the change?

Thanks for your patience on this one.

joeldenning:
In the build you referenced above, it looks like you might not have enabled the native-iterator tests in the travis.yml?? In order to test if Evergreen -> Windows 7 works, those tests must be enabled. My latest commit changes both the travis.yml and the karma.conf -- we'll see if it works.

joeldenning:
Looks like switching to Windows 7 may not have worked. See https://travis-ci.org/ModuleLoader/es6-module-loader/jobs/88651661

guybedford:
@joeldenning I finally managed to get this to pass in https://travis-ci.org/ModuleLoader/es6-module-loader/jobs/89482001#L920, by trying every combination I could. Let me know if that seems like it cracks it?

joeldenning:
I merged your changes in and now it is have passing. I'm not sure what is the difference between travis-ci/pr vs travis-ci/push, but I think it's ready to merge? Let me now if you'd like me to squash commits or do something to get the travis-ci/pr check to succeed.

guybedford:
Great, I think that's just because the Saucelabs config isn't set up your side.

PR #441, opened 2015-10-13
jsdom support in baseURI recognition

pawelgalazka:
https://github.com/systemjs/systemjs/issues/840


guybedford:
I'm not sure this is the correct fix. I've included my comments on the original issue.

PR #442, opened 2015-10-13
new System.noConflict() function to restore state of the global System

drieks:
Hi,

here is a little function you can use to restore the previous state of the global System var if you are using more than one System Instance.


guybedford:
Thanks for posting. I agree it makes sense to have a more cautious polyfill option and have created an issue for this in https://github.com/ModuleLoader/es6-module-loader/issues/443. I don't want to extend the API though as we need to adhere to the spec.

PR #446, opened 2015-11-06
Removing Loader.prototype.install and Loader.prototype.lookup and put…

joeldenning:
…ting them on the Registry.prototype

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -116,63 +116,6 @@
       throw new TypeError('Invalid stage ' + stage);
   };
 
-  // TODO: This has been moved to Registry.prototype.lookup (4.4.3)
-  Loader.prototype.lookup = function(key) {
-    var loader = this._loader;
-
-    var entry = loader.registry[key];
-    if (!entry)
-      return null;
-
-    var state;
-    if (entry.state == FETCH)
-      state = 'fetch';
-    else if (entry.state == TRANSLATE)
-      state = 'translate';
-    else if (entry.state == INSTANTIATE)
-      state = 'instantiate';
-    else if (entry.state == LINK)
-      state = 'link';
-    else if (entry.state == READY)
-      state = 'ready';
-
-    return {
-      state: state,
-      metadata: entry.metadata,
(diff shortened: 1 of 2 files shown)


joeldenning:
The failing tests seem to be related to the same problem in my last pr #439. Somehow travis-ci/pr is different than travis-ci/push. Let me know if there is something I need to do to make my pr's have the correct travis-ci configuration.

src/loader.js
```diff
@@ -253,6 +225,35 @@
...
+  };
+
+  // 4.4.4
+  Registry.prototype.install = function(key, module) {
+    if (typeof this !== 'object')
+      throw new TypeError('registry must be an object');
+    if (this._registry.registryData[key])
+      throw new TypeError('Module with key ' + key + ' already exists');
+    if (!(module instanceof Module))
+      throw new TypeError('module must be an instance of Module');
```
guybedford: Is this check in the spec?

joeldenning: The checks on lines 235 and 237 are in the spec, but the check on line 239 is not. The reason I put it in here was because the same check was in the Loader.prototype.install function and it seemed to make sense. However, as I think about that reasoning, it's probably best just to follow the spec exactly so I can change this.

For my own reference later on, http://whatwg.github.io/loader/#registry-prototype-install

guybedford:
That was quick! Thanks, it's really nice to be moving on these updates, as it's not something I have time to personally work on right now.

I just checked and this passed fine on https://travis-ci.org/ModuleLoader/es6-module-loader/builds/89703716.

In terms of spec-compliancy, note that this code will be changing in https://github.com/whatwg/loader/pull/97. I'm happy to update this in due course when there is progress there though.

I've just added you to the organization here so next time you can run PRs against local branches which should ensure the tests pass as well.

guybedford:
Actually were you going to look at that `module instanceof Module` thing?

joeldenning:
Yeah I was just deleting the `module instanceof Module` check since it's not in the spec. I can make that a separate pr, though

Issue #374, opened 2015-05-11
Support loading in a Service Worker


matthewp:
I think this would work if https://github.com/ModuleLoader/es6-module-loader/blob/master/src/system.js#L64 checked for global.fetch and used that.

guybedford:
Does XHR not work inside of service workers?

OliverJAsh:
How would we download and execute SystemJS in a service worker at all?

matthewp:
@guybedford I believe not, it's hitting the `throw` in this case.  Eventually all environments will support `fetch` so it would be nice to make that the default anyways.

@OliverJAsh You can execute any script as a service worker, so that includes es6ml.  You might want to bundle it with some other script that calls `System.import` if you want it to actually do something though.

guybedford:
Ok, sure, yes it makes sense to use `fetch` if its supported as an API here anyway.

guybedford:
I believe support for this was added in https://github.com/ModuleLoader/es6-module-loader/pull/448.

PR #448, opened 2015-11-09
Add support for new async fetch API

jgraichen:
This commit adds a `fetchTextFromURL` strategy using the new fetch API required for service workers. #374

It can be used in a service worker like this:

```js
importScripts('babel-browser.js')
importScripts('es6-module-loader-dev.js')

self.addEventListener('install', function(event) {
    event.waitUntil(System.import('sw.jsx').then(function(sw) {
        return sw.install(event);
    }));
});

// [..]
```

I did not get a reliable test case working registering and unregistering a service worker due to mismatching URL scopes in test runners.

### DIFF
--- a/src/system-fetch.js
+++ b/src/system-fetch.js
@@ -59,7 +59,7 @@
       }
     };
   }
-  else if (typeof require != 'undefined') {
+  else if (typeof require != 'undefined' && typeof process != 'undefined') {
     var fs;
     fetchTextFromURL = function(url, authorization, fulfill, reject) {
       if (url.substr(0, 8) != 'file:///')
@@ -83,6 +83,29 @@
         }
       });
     };
+  }
+  else if (typeof self != 'undefined' && typeof self.fetch != 'undefined') {
+    fetchTextFromURL = function(url, authorization, fulfill, reject) {
+      var opts = {
+        headers: {'Accept': 'application/x-es-module, */*'}
+      };
+
+      if (authorization) {
+        if (typeof authorization == 'string')
+          opts.headers['Authorization'] = authorization;
+        opts.credentials = 'include';
+      }
+
+      fetch(url, opts)
+        .then(function (r) {
+          if (r.ok) {
+            return r.text();
+          } else {
+            throw new Error('Fetch error: ' + r.status + ' ' + r.statusText);
+          }
+        })
+        .then(fulfill, reject);
+    }
   }
   else {
     throw new TypeError('No environment fetch API available.');


src/system-fetch.js
```diff
@@ -1,5 +1,31 @@
   var fetchTextFromURL;
-  if (typeof XMLHttpRequest != 'undefined') {
+  if (typeof fetch != 'undefined') {
```
guybedford: Can we do this check after the `XMLHttpRequest` check to ensure it is only used when needed?

src/system-fetch.js
```diff
@@ -1,5 +1,31 @@
...
+    fetchTextFromURL = function(url, authorization, fulfill, reject) {
+      var opts = {
+        headers: {'Accept': 'application/x-es-module, */*'}
+      };
+
+      if(authorization) {
```
guybedford: `if (authorization)`

src/system-fetch.js
```diff
@@ -1,5 +1,31 @@
...
+            return r.text();
+          } else {
+            throw new Error('Fetch error: ' + r.status + ' ' + r.statusText);
+          }
+        })
+        .then(function (r) {
+          fulfill(r);
+        }).catch(function (err) {
+          reject(err);
+        });
```
guybedford: This can just be `.then(fulfill, reject)` equivalently.

guybedford:
:+1:

jgraichen:
I've updated and moved the code. Thanks for comments.

src/system-fetch.js
```diff
@@ -59,6 +59,29 @@
       }
     };
   }
+  else if (typeof fetch != 'undefined') {
```
guybedford: I'm just trying to think carefully about the environments here - if I had `fetch` defined as a global in Node, I probably wouldn't want it to be used either right? So perhaps we should drop it below the `typeof require != 'undefined'` check as well? Then the issue is if we're in a service worker with `require` defined, so perhaps we can add a `typeof process != 'undefined'` check as well to avoid that.

jgraichen: `fetch` isn't actually specific to service workers but the "new" [promise-based request API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) that is intended to replace XMLHttpRequest for browser windows and web workers. Service worker are special workers in that case as they explicit do not allow the older XHR methods.

Maybe one could test for `self.fetch` as `self` is the `Window`, `WorkerGlobalScope` or `ServiceWorkerGlobalScope` object at least in a modern Chrome. It is still nothing that cannot be overridden.

If using `fetch` should be restricted to only service workers (even as it works in windows and web workers) something like this may work ([source](https://stackoverflow.com/a/18002694)):

```
if (typeof ServiceWorkerGlobalScope !== 'undefined' && self instanceof ServiceWorkerGlobalScope)
```

guybedford: I'm not saying it should be restricted to service workers, I'm just saying we shouldn't default to using it in Node, where we want to use `fs` not HTTP for loading.

jgraichen: Yes, of course. I'm just unsure what a good way to check the environment would be.

I'll move it below the require check and add a process check there.

guybedford:
Great, thanks for this. Looks good to me and I will trust is working ok for your use cases.

PR #450, opened 2015-11-15
Registry functions -- uninstall and cancel

joeldenning:


joeldenning:
I was hoping the travis-ci tests would pass without creating a new branch on this repo, but that does not appear to be the case. I'm closing this in favor of #451

PR #460, opened 2016-01-01
Implemented 3.3.5 - the getter for a loader's registry.

joeldenning:


guybedford:
:+1: 

btw I'm thinking we should publish the new spec work together as soon as possible and publish it as a 0.20. Even though SystemJS uses the old spec, there's no reason to hold this project back anymore when we have the loader working to the new spec already.

joeldenning:
Interesting, I'll try to keep the pull requests coming then :). I've started work on the refactor that caridy committed a while ago to Registry objects and ModuleStatus objects, but am wondering what to do about #461 . What are your thoughts on using Map?

guybedford:
Excellent thanks so much for your work here, that refactor would be really great to get this project released to parity.

Issue #461, opened 2016-01-01
Should Map be used internally?


joeldenning:
The latest commits by caridy in the spec start using Maps internally and also the Registry objects now have 8 functions that are pass-throughs to Map functions. Maps, of course, are not fully supported in any version of IE but are supported in the other major browsers.

What are your thoughts on depending on the es6-map polyfill at https://www.npmjs.com/package/es6-map? If not, then we should polyfill the following functions to act exactly like their `Map` counterparts?
```
Registry.prototype[Symbol.iterator]
Registry.prototype.entries
Registry.prototype.keys
Registry.prototype.values
Registry.prototype.get
Registry.prototype.set
Registry.prototype.has
Registry.prototype.delete
```

guybedford:
Code size is a big concern for the loader, as it is part of the critical load path itself.

Looking at the polyfill implementation it seems to use quite a lot of dependencies etc.

The get, set, has, delete, keys and values outputs seem straightforward?

If it's possible to write the iterator and entries in a flexible way that can work in ES5, ES5 + polyfill, and ES6 I think that would be ideal, but I completely understand if it's a tough ask.

joeldenning:
I'll give it a go

PR #462, opened 2016-01-03
Implementing 4.1.1 through 4.4.9

joeldenning:
One of the big things to call out is the newly introduced dependency on es6-map. See #461 for the discussion on pros/cons of that decision.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -186,14 +186,14 @@
     if (__global.Map) {
       // native iterator
       return this._registry.registryMap.entries();
-    } else {
+    } else if (__global.Symbol && __global.Symbol.iterator) {
       // polyfilled iterator
       var keys = Object.keys(this._registry.registryMap);
       var keyIndex = 0;
       var instance = this;
       return {
         next: function() {
-          if (keyIndex < keys.length - 1) {
+          if (keyIndex < keys.length) {
             return {
               value: [keys[keyIndex], instance._registry.registryMap[keys[keyIndex++]]],
               done: false
@@ -206,6 +206,8 @@
           }
         }
       };
+    } else {
+      throw new Error('Symbol.iterator must be polyfilled to iterate over registry entries');
     }
   }
 
@@ -216,13 +218,13 @@
     if (__global.Map) {
       // native iterator
       return this._registry.registryMap.keys();
-    } else {
+    } else if (__global.Symbol && __global.Symbol.iterator) {
       // polyfilled iterator
       var keys = Object.keys(this._registry.registryMap);
       var keyIndex = 0;
       return {
         next: function() {
-          if (keyIndex < keys.length - 1) {
+          if (keyIndex < keys.length) {
             return {
               value: keys[keyIndex++],
               done: false
@@ -235,6 +237,8 @@
           }
         }
       };
+    } else {
+      throw new Error("Symbol.iterator must be polyfilled in order to iterate over registry keys")
     }
   }
 
@@ -245,14 +249,14 @@
     if (__global.Map) {
       // native iterator
       return this._registry.registryMap.values();
-    } else {
+    } else if (__global.Symbol && __global.Symbol.iterator) {
       // polyfilled iterator
       var keys = Object.keys(this._registry.registryMap);
       var keyIndex = 0;
       var instance = this;
       return {
         next: function() {
-          if (keyIndex < keys.length - 1) {
+          if (keyIndex < keys.length) {
             return {
               value: instance._registry.registryMap[keys[keyIndex++]],
               done: false
@@ -265,6 +269,8 @@
           }
         }
       };
+    } else {
+      throw new Error("Symbol.iterator must be polyfilled in order to iterate over registry values");
     }
   }
 
--- a/src/loader.js
+++ b/src/loader.js
@@ -153,13 +153,16 @@
   function createRegistry() {
     var registry = new RegistryPrototype();
     registry._registry = {};
-    if (__global.Map)
+    var iterator;
+    if (__global.Map && __global.Map.prototype && __global.Map.prototype.entries && __global.Map.prototype.keys
+         && __global.Map.prototype.values && __global.Map.prototype.get && __global.Map.prototype.set
+         && __global.Map.prototype.has && __global.Map.prototype.delete)
       registry._registry.registryMap = new __global.Map();
     else
-      registry._registry.registryMap = {};
+      registry._registry.registryMap = new InternalMapPolyfill();
     // 4.4.2
     if (__global.Symbol && __global.Symbol.iterator)
-      registry[__global.Symbol.iterator] = RegistryPrototype.prototype.entries;
+      registry[__global.Symbol.iterator] = mapPolyfillEntriesIterator.bind(registry._registry.registryMap);
     return registry;
   }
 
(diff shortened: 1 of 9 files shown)


joeldenning:
As you can see, I'm struggling to get the tests to pass because Map is not defined natively in the browser and for some reason core-js is not polyfilling Map like I thought it would be. I'll take a further look at this tomorrow.

joeldenning:
Looks like it actually is working fine except for in a WorkerGlobalScope. Will think about it more and see if their is a way to get the polyfill to add Maps to the WorkerGlobalScope, in addition to the window. If not, we may have to re-polyfill Maps and MapIterators inside of es6-module-loader

joeldenning:
Per #461, I'll remove the dependency on es6-map and polyfill it myself. That should also solve the web worker problem that I was describing above.

guybedford:
@joeldenning sounds great. Do let me know if you think the DIY polyfill interface approach isn't going to work though. I only just checked and the map polyfill is a mere 6KB minfied and gzipped, so it certainly isn't ruled out.

probins:
would this DIY polyfill be always loaded, even in the majority of browsers that support natively? I would prefer it if all polyfilling were done in userland. Current support for Map is similar to that for Promise. I currently check for Promise support and load polyfill if necessary before loading the module loader, and could easily add something similar for Map. I don't think the loader should be bloated with code just to support out-of-date browsers.

guybedford:
@probins the idea is not a polyfill per se, just a small compatible API for Map objects on the exposed interface from the loader.

test/registry.spec.js
```diff
@@ -1,183 +1,303 @@
...
+    });
 
+    it('should fail if called with new', function() {
+      expect(function() { new Registry() }).to.throwException();
     });
 
   });
 
-  describe('install function', function() {
+  describeIf(System.transpiler !== 'traceur', 'entries function', function() {
```
joeldenning: traceur thinks that any variable that you call `next` on is a generator and then errors out because an iterator isn't really a generator

test/registry.spec.js
```diff
@@ -1,183 +1,303 @@
...
+      expect(function() {Registry.prototype.entries.call('not a real registry')})
+      .to.throwException(function(e) {
+        expect(e).to.be.a(TypeError);
       });
     });
 
   });
 
-  describe('lookup function', function() {
+  describeIf(System.transpiler !== 'traceur', 'keys function', function() {
```
joeldenning: same here

test/registry.spec.js
```diff
@@ -1,183 +1,303 @@
...
-    it('throws if the registry data is corrupted', function() {
-      registry._registry.registryData['key'] = "string that's not an entry";
-      expect(function() { registry.lookup('key') }).to.throwException();
-    });
+  });
 
-    it('returns the module if it is in a ready state', function() {
-      var module = new ModulePolyfill({});
-      registry.install('key', module);
+  describeIf(System.transpiler !== 'traceur', 'values function', function() {
```
joeldenning: same here

joeldenning:
I implemented everything without the es6-map dependency. The functions on the Registry that return iterators will throw if the browser does not support (nor has been polyfilled to support) iterators. This should not affect existing users of es6-module-loader since System.registry.entries() / keys() / values() are new apis that have been introduced recently into the spec

test/registry.spec.js
joeldenning: See https://travis-ci.org/ModuleLoader/es6-module-loader/builds/101356889 and https://travis-ci.org/ModuleLoader/es6-module-loader/jobs/101356890#L969

src/loader.js
```diff
@@ -146,15 +147,186 @@
...
 
   // 4. Registry Objects
+  // For now, registry objects are a work in progress that don't fully integrate into the rest of the code base
 
   // 4.1.1
+  function createRegistry() {
+    var registry = new RegistryPrototype();
+    registry._registry = {};
+    if (__global.Map)
+      registry._registry.registryMap = new __global.Map();
```
guybedford: Are there any browsers that provide incomplete or non-functional implementations of Map? Or is it always safe to assume that everything else will work correctly here?

src/loader.js
```diff
@@ -146,15 +147,186 @@
...
+    if (typeof this !== 'object')
+      throw new TypeError('cannot get entries of a non-registry');
+    if (__global.Map) {
+      // native iterator
+      return this._registry.registryMap.entries();
+    } else if (__global.Symbol && __global.Symbol.iterator) {
+      // polyfilled iterator
+      var keys = Object.keys(this._registry.registryMap);
+      var keyIndex = 0;
+      var instance = this;
```
guybedford: We should probably just store `registryMap` instead of `this` here?

src/loader.js
```diff
@@ -146,15 +147,186 @@
...
+  // 4.4.2 is inlined in 4.1.1
+
+  // 4.4.3
+  RegistryPrototype.prototype.entries = function() {
+    if (typeof this !== 'object')
+      throw new TypeError('cannot get entries of a non-registry');
+    if (__global.Map) {
+      // native iterator
+      return this._registry.registryMap.entries();
+    } else if (__global.Symbol && __global.Symbol.iterator) {
```
guybedford: If the user is using a polyfill for iterator that does not overwrite the global `Symbol.iterator`, surely it should still be possible for them to use this `entries` feature though? That is, do we really need this check?

src/loader.js
```diff
@@ -146,15 +147,186 @@
...
+              value: undefined,
+              done: true
+            };
+          }
+        }
+      };
+    } else {
+      throw new Error("Symbol.iterator must be polyfilled in order to iterate over registry values");
+    }
+  }
```
guybedford: With all of these functions above - `values`, `keys`, `entries`, I'm wondering if we shouldn't just create a `InternalMapPolyfill` class that we can instantiate in 4.1.1 rather?

guybedford:
:+1: looks really great, and the polyfill overhead doesn't seem too bad at all. Are there any other limitations of the polyfill approach?

src/loader.js
joeldenning: Good point. According to https://kangax.github.io/compat-table/es6/, All versions of Safari are okay because they either don't support maps at all (safari 7-) or they support everything we need (safari 8+). However, IE11 is not okay because it supports the Map constructor but not the `entries`, `keys`, or `values` functions which we need. As for Chrome and Firefox, the table doesn't go back far enough to show whether there was ever a period where Map was supported without all of the functions we need on its prototype.

I'll update this so that we feature check for all of the functions that we need on the Map prototype.

src/loader.js
joeldenning: Probably a good idea -- especially if we need to use Maps elsewhere. Will change.

src/loader.js
joeldenning: Done, as a side effect of switching over to using `InternalMapPolyfill`

src/loader.js
joeldenning: I don't think I understand -- the entries function is required to return an iterable object, where iterable objects have a `Symbol.iterator` property. How would a user polyfill for iterator without overwriting/using `Symbol.iterator`?? What other property would the polyfill look at when trying to get the values of an iterable???

Perhaps you're recommending returning something that is not an iterable at all??

joeldenning:
Some things to note in the latest commit:

I created `InternalMapPolyfill` as you suggested. Also, I found some bugs in my implementations of `values`, `entries`, and `keys` on the registry where they were not really returning iterables that you can `for ... of` over. The reason why those bugs were not caught in the previous round of testing is because I only ran the `for ... of` tests when using a native Map (instead of the polyfill). I have improved the tests by adding to the travis config some more combinations: we now run the `for ... of` tests (--native-iterator) after having ensured that `InternalMapPolyfill` is being used instead of the native Map.

Also, you can see my response to your comments on the previous revision -- looks like the last remaining discussion point is related to whether to `throw` or not, when `Symbol.iterator` is not defined

joeldenning:
Oh and in terms of limitations to `InternalMapPolyfill`, I think all of the functions that are implemented actually do a decent job of implementing the spec. However, I have not implemented `InternalMapPolyfill.prototype.forEach` nor `InternalMapPolyfill.prototype.size` since they were not needed for `Registry` objects. Other than those functions, I'm not aware of any limitations (although admittedly I have not read through the entire `Map` spec)

joeldenning:
ping

src/loader.js
```diff
@@ -146,15 +147,99 @@
...
+    // 4.4.2
+    if (__global.Symbol && __global.Symbol.iterator)
+      registry[__global.Symbol.iterator] = mapPolyfillEntriesIterator.bind(registry._registry.registryMap);
+    return registry;
+  }
+
+  // 4.2
+  function Registry() {
+    throw new Error('A registry may only be created when creating a loader');
+  }
```
guybedford: This may be a little too strict. This would only be necessary if the Registry constructor was public, which I don't believe it is. Since this is the case, simply using an internal `Registry` class is fine since it will only be instantiated within 4.1.1 surely?

src/loader.js
```diff
@@ -146,15 +147,99 @@
   };
 
   // 4. Registry Objects
+  // For now, registry objects are a work in progress that don't fully integrate into the rest of the code base
 
   // 4.1.1
+  function createRegistry() {
+    var registry = new RegistryPrototype();
+    registry._registry = {};
```
guybedford: Should this not be done in the `Registry` constructor? Does it correspond to a step?

src/loader.js
```diff
@@ -146,15 +147,99 @@
...
+    var iterator;
```
guybedford: This doesn't seem to be used at all?

src/loader.js
```diff
@@ -146,15 +147,99 @@
...
+    if (__global.Map && __global.Map.prototype && __global.Map.prototype.entries && __global.Map.prototype.keys
+         && __global.Map.prototype.values && __global.Map.prototype.get && __global.Map.prototype.set
+         && __global.Map.prototype.has && __global.Map.prototype.delete)
```
guybedford: Rather than inlining this check here, perhaps we can move it to the execution scope outside of this function, setting a `hasNativeMap` variable, which can then just be checked here? It might be a clearer separation of concerns as well.

guybedford:
Thanks for your patience, I will aim to be quicker with these reviews.

src/loader.js
joeldenning: I was under the impression that the Registry constructor actually would be public because of the language in https://whatwg.github.io/loader/#registry-constructor. The reason for that is because it talks about the Registry constructor as a separate thing from the %Registry% intrinsic object. Admittedly, though, I don't have any real training / experience on how to read these specs and so maybe I am misinterpreting. Could you shed some light on how the verbage in that link should be interpreted?

src/loader.js
joeldenning: This corresponds to https://whatwg.github.io/loader/#registry-CreateRegistry. However, depending on whether `Registry` is public or not, we could potentially take some shortcuts that deviate from the spec but make the code smaller/simpler

src/loader.js
joeldenning: I like it -- done.

src/loader.js
joeldenning: Good catch -- removed.

src/loader.js
guybedford: I think the wording there that it isn't meant to be instantiated indicates that it is an internal class, but it is yet to be determined how public the API is I believe. If it were to become public, the ability to run the constructor of the registry would be supported. Either way I think having the Registry as a standard class here would be close enough to the spec. It's worth keeping the implementation simple where possible I think, complexity can always be added later.

src/loader.js
joeldenning: Unless you were talking about line 155 specifically, where `registry._registry` is being created?  That line was intended to implement steps 2-3 of https://whatwg.github.io/loader/#registry-CreateRegistry. The reason for using a `_registry` property was to follow the pattern I had seen elsewhere about underscore-prefixed properties that are "internal" (see [example](https://github.com/ModuleLoader/es6-module-loader/blob/1.0/src/loader.js#L13)). 

However, now seeing your other comment about the above diff, it looks like we're planning on deviating a bit from the spec in the name of simplicity. So in that case, I can just put all of this into the `Registry` constructor.

src/loader.js
joeldenning: Fair enough, will do. For future reference, how does one know which properties in the spec are public or not? For example, I know the `System.loader` will be public, but I'm not sure how the spec indicates that that one will be public but Registry will not. See https://whatwg.github.io/loader/#system-loader-instance for reference

guybedford: Yes exactly, I believe which intrinsics are set as globals is determined by if they are exported as globals in the browser spec part there.

joeldenning:
Just addressed all of your latest comments in these last two commits. Did you see my comment about throwing if there is not native iterator support inside of the `entries`, `keys`, and `values` functions? Did you have any more thoughts on that?

src/loader.js
```diff
@@ -146,15 +153,89 @@
   };
 
   // 4. Registry Objects
+  // For now, registry objects are a work in progress that don't fully integrate into the rest of the code base
+
+  // 4.1.1 inlined in 4.2
+
+  // 4.2 - see https://github.com/ModuleLoader/es6-module-loader/pull/462#discussion-diff-50639828 for why it deviates from spec
+  function Registry() {
+    this._registry = {};
```
guybedford: I think we can set `this.registryMap` directly here. The use of `_loader` on Loader is done to mask the private class state on a public class. If this was public, I'd simply rename `this.registryMap` to `this._registryMap`. Ideally in modern environments we'd just use private Symbols here.

joeldenning: done

guybedford:
Throwing on the entries, keys and values functions when no iterator is present seems fine to me, although if we do need to do these registry iterations internally, we should make sure we don't use these paths then.

:+1: do go ahead with the merge.

PR #464, opened 2016-01-22
addToError use original error object

dmail:
Related to https://github.com/ModuleLoader/es6-module-loader/issues/399 and https://github.com/ModuleLoader/es6-module-loader/pull/404.

From what I understood, `addToError` function (see [wrapper-start.js#L39](https://github.com/ModuleLoader/es6-module-loader/blob/master/src/wrapper-start.js#L39)) is responsible to append a contextual message to give more detail about where an error hapenned.

But I think there is an issue with the current implementation, imagine the following module named `error.js`

```javascript
var typeError = new TypeError('foo is not defined');
throw typeError;
```

And a file importing it

```javascript
System.import('./error.js').catch(function(e){
  console.log(e.name); // e.name is "Error" instead of "TypeError"
});
```
`e.name == 'Error'`because we received error created inside the `addToError` function and not the `typeError` object thrown in `error.js`.

### DIFF
--- a/src/wrapper-start.js
+++ b/src/wrapper-start.js
@@ -34,24 +34,15 @@
   })();
 
   function addToError(err, msg) {
-    var newErr;
+
     if (err instanceof Error) {
-      newErr = new Error(err.message, err.fileName, err.lineNumber);
-      if (isBrowser) {
-        newErr.message = err.message + '\n\t' + msg;
-        newErr.stack = err.stack;
-      }
-      else {
-        // node errors only look correct with the stack modified
-        newErr.message = err.message;
-        newErr.stack = err.stack + '\n\t' + msg;
-      }
+      err.message+= '\n\t' + msg;
     }
     else {
-      newErr = err + '\n\t' + msg;
+      err+= '\n\t' + msg;
     }
-      
-    return newErr;
+
+    return err;
   }
 
   function __eval(source, debugName, context) {


guybedford:
Thanks. We actually used to do this until it was changed in https://github.com/ModuleLoader/es6-module-loader/commit/7f77b6a8e62c050434c6955abdc2385af2b9b007. I think the idea was that by mutating the original error, anything else catching that same original error would be getting a mutated object.

Is it just for ensuring the error type that this is needed?

dmail:
The reason is that, on v8 (nodejs), doing `error.stack = something` prevent Error.prepareStackTrace from being called and breaks a module like [node-source-map-support](https://github.com/evanw/node-source-map-support).

In the same move I realized that `addToError` creates an other error than the original one, loosing error type and any other own property of it (some may want to add a .code property as nodejs does with ENOENT).

guybedford:
@dmail is it not possible to patch that stack loss with a call to `captureStackTrace`? I've been meaning to line this implementation up closer to https://github.com/davepacheco/node-verror/blob/master/lib/verror.js#L91 which seems to do this well (and losing the class name is a natural consequence there too).

dmail:
@guybedford It seems `captureStackTrace` will not help much the reason is line [#47](https://github.com/ModuleLoader/es6-module-loader/blob/master/src/wrapper-start.js#L47) where `newErr.stack = err.stack + '\n\t' + msg;`.

I see in the source code `// node errors only look correct with the stack modified` but to me it's perfectly ok to do `newErr.message = error.message + '\n\t' + msg;` on node and let stack property untouched.

addToError is forming an error chain, so could we just add a `cause` property to newErr letting consumer of your library aware of the error chain just like vError does ?

guybedford:
@dmail do you think the `captureStackTrace` would avoid the need to do the `err.stack` mutation? This was purely so that `console.log(err)` would look nice and include the full stack in Node so it may well be the way to go to avoid the stack adjustment?

dmail:
@guybedford If I get it right `captureStackTrace` make error.stack think the error was generated from a specific location. It cannot simulate an error chain or add contextual information about an existing error.

I can modify this pull request later to propose you something closer to vError and that suits what you want to do.

guybedford:
Ok right I guess so, but it still sounds better than mutating the stack. Yes I'd like to align with the vError approach entirely, and that sounds like it would avoid the need for stack modification. A PR would be a great help, this has been on the list for a while.

dmail:
I just want to inform you that I will just mirror what vError already does : create an error, mutate its message property & register a cause on it.

It will keep stack untouched, allow error chain discovery and `console.log(err)` will be ok.

guybedford:
@dmail it may actually be best not to waste your time here and for me to do this, as it would be good to test out all the cases carefully.

dmail:
@guybedford Ok let's do this because you have a better vision of the implications. I hope to see this soon thank you very much for your attention.

guybedford:
I finally got to this in https://github.com/ModuleLoader/es6-module-loader/commit/dddb5223b5ff19c22a5bdfbdd3ed5ed12d4b58aa. Amazing how simple it is in retrospect... where are the good blog posts on this topic! The errors look much nicer now, thanks for your patience and input here @dmail.

guybedford:
I think we can retain the original error at `e.originalError` rather to solve this use case. Added in https://github.com/ModuleLoader/es6-module-loader/commit/d78178d27a2a0618089f28031ac85cdcf194565b.

guybedford:
After all that I've reverted to the error mutation approach. The reason being that it allows us to inject the SystemJS error information without having to do stack modifications, resulting in the error displaying equivalently between Node and the browser. It's a much simpler mechanic with less things that can go wrong, and correctly preserves the original error stack with debug points from the first error. It's not ideal to be mutating the error message, but this injection kind of makes sense if you imagine SystemJS as owning the error process for errors thrown within code it runs.

dmail:
As long as there is `e.originalError`, it's ok ;)

PR #469, opened 2016-01-25
Adding preprocessor step to remove out strict correctness in producti…

### LINKED ISSUE #467: strict v non-strict builds
Since we're implementing to the spec based on correctness, it would be nice to have a globally-scoped variable, `STRICT`, which is included as a boolean check whenever doing checks that one wouldn't want in a high-performance production loader where correctness is not important.

We can then run two builds of the loader through dead code removal that set STRICT=false and STRICT=true respectively to get different builds for the different requirements here.

joeldenning:
…on environments. Fixes #467

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -20,7 +20,7 @@ module.exports = function (grunt) {
     concat: {
       dist: {
         files: {
-          'dist/<%= pkg.name %>.src.js': [
+          'dist/<%= pkg.name %>-strict.src.js': [
             'src/wrapper-start.js',
             'src/loader.js',
             'src/dynamic-only.js',
@@ -28,7 +28,7 @@ module.exports = function (grunt) {
             'src/system.js',
             'src/wrapper-end.js'
           ],
-          'dist/<%= pkg.name %>-dev.src.js': [
+          'dist/<%= pkg.name %>-dev-strict.src.js': [
             'src/wrapper-start.js',
             'src/loader.js',
             'src/declarative.js',
@@ -38,6 +38,17 @@ module.exports = function (grunt) {
             'src/module-tag.js',
             'src/wrapper-end.js'
           ]
+        }
+      }
+    },
+    preprocess: {
+      multifile: {
+        files: {
+          'dist/es6-module-loader-dev.src.js': 'dist/es6-module-loader-dev-strict.src.js',
+          'dist/es6-module-loader.src.js': 'dist/es6-module-loader-strict.src.js'
+        },
+        context: {
+          STRICT: true
         }
       }
     },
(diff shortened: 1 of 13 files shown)


joeldenning:
I recommend pulling down the branch, running `grunt`, and then checking the `.src.js` and the minified files for the difference between the strict and non-strict built files.

Gruntfile.js
```diff
@@ -28,7 +28,7 @@ module.exports = function (grunt) {
             'src/system.js',
             'src/wrapper-end.js'
           ],
-          'dist/<%= pkg.name %>-dev.src.js': [
+          'dist/<%= pkg.name %>-dev-strict.src.js': [
```
guybedford: I'm just wondering, since we already separately do the `dev` and `production` builds, where the `dev` build is the one that provides the declarative linking layer (which is then left out of the non-dev build), perhaps we just need to keep it as these two builds and rename `dev` to `strict`. Or do you think it's important to have the non-dev strict build for compatibility?

`dev` is also no longer the right name for the other build as it is no longer used for dev workflows. The best name I can come up with right now is `declarative`?

src/loader.js
```diff
@@ -88,22 +88,26 @@
         return entry.module;
       });
 
+    // @ifdef STRICT
```
guybedford: A custom syntax worries me a little as it ties us to a specific pre-processor. While I know a global if statement is uglier, the nice thing is that it will be compatible with any preprocessor tooling in future avoiding locking us into a specific approach.

Gruntfile.js
joeldenning: Yeah that makes sense. I'll change over to `declarative` instead of `dev`.

src/loader.js
joeldenning: Yeah I started out with uglifyjs' preprocessor removing out the strict comparison stuff. But then I realized that since we don't run uglify on the `.src.js` file that uglify wouldn't be a good option for [`es6-module-loader.src.js`](https://github.com/ModuleLoader/es6-module-loader/blob/master/dist/es6-module-loader.src.js), but only for `es6-module-loader.js`.

I'll look into maybe some other options that will preprocess out the strict code that becomes dead when the global variable is set.

joeldenning: @guybedford I looked for preprocessors that were not comment based and really only found uglifyjs or babel-dead-code-elimination (which seems to maybe be just a babel 5 thing that is not anymore supported and has a bad github link).

Do you have any ideas for a variable/code based preprocessor that removes the dead code during the preprocessor step? If not, would you prefer the uglifyjs route (where es6-module-loader.src still has all the strict correctness code in it and is the same as es6-module-loader-declarative.src), or would you prefer the comment-based approach that I've already implemented?

guybedford: I guess let's stick with this method for now then actually and we can easily re-evaluate in future.

joeldenning:
Just switched over to having just to built versions -- declarative and production.

guybedford:
Great, thanks.

PR #478, opened 2016-02-29
Report errors to console that bubble up during linking

jryans:
This issue was reproduced with SystemJS 0.19.21 which uses es6-module-loader 0.17.10.

I've noticed that some valid errors are currently reported sub-optimally to the console, giving only the "main" message without a full stack, as in:

truncated-errors

If we instead use `console.error` to report these, you get a full stack trace and much more useful information:

full-stack

I am not sure if this is the right place to add such logging, so there may easily be a better way to add this kind of logging. In any case, this has helped me a lot when debugging errors.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -473,11 +473,13 @@ function logloads(loads) {
     try {
       link(linkSet, function(load, exc) {
         linkSetFailed(linkSet, load, exc);
+        console.error(exc);
         error = true;
       });
     }
     catch(e) {
       linkSetFailed(linkSet, null, e);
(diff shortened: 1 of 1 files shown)


guybedford:
If you catch the error from the top-level System.import promise, can you not do this `console.error` in there rather? We shouldn't enforce console output ever.

jryans:
Ah, sure enough, you can. That does indeed seem much better, thanks!

I may submit a PR to SystemJS docs to emphasize this point, as I don't think I realized `import` returned a promise.

PR #480, opened 2016-03-13
System.module() enhancement

joebrassi:
This Pull Request contains three fixes:

1.-  Adding a _name_ to the modules loaded by `System.module()`. This name is useful for subsequent identification of the module (i.e. Avoid multiple loading of the same module, allow module deletions, etc).
Example:
```
function buttonClick() {
  // dynamic loading API
  var source = `
export class Foo {
   constructor() {
      console.log('Created the ES6 class foo!');
   }
}
`;
  var options = {
    name: 'someName'
  };
  System.module(source, options).then(function(module) {
    new module.Foo();
  });
}
```
2.-  As all the modules has a _name_, the `System.module()` was failing if an already-registered module was being loaded by this command. So it now checks whether the module exists to proceed with the loading.
3.-  The System.load() was not working, or I didn't understand how was it working. It now returns the module requested by its name.
Example:
```
function button1Click() {
  System.load('someName').then(function(module) {
    new module.Foo();
  })
}
```


guybedford:
I think `System.define` is already serving this purpose, which takes `name, source` as its arguments.

Note that `System.load` is a non-executing import based on a normalized module name. Usually System.import is preferred.

joebrassi:
If `System.define` does the same as `System.module`, then why you have two methods that perform almost the same task? Anyways, I think that `System.module` do what it has to do, that is to create, transpile and return the module as a Promise, but if `System.define` does the same, could you please share a demo or some sort of documentation about how it works? I asked here some time ago for such functionality and got no satisfactory response, so that's why I went to the code and wrote a couple of lines which I am sharing here.

About the `System.load`, again could you please share a demo or some documentation about how to load an -already existent- module by _name_ with `System.import`? So far I've seen it working using a `path` to the `.ts` file, which it seems to download, transpile, etc, etc.

probins:
`System.module()` is for anonymous modules, the equivalent of `<script type="module">...</script>`. Anonymous modules don't have names.

This polyfill is following the spec. The code does what it does because that's what the spec says it should do. If you want to avoid duplicate loads, you should use `System.import()`; the demo in this repo is _very_ basic, but there is an example of this there: if you click on the button twice, it does not reload the module.

PR #481, opened 2016-03-19
remove options.modules in babelTranspile

Treri:
options.modules = 'system' is deprecated in Babel 6

And this will lead to error 

> Using removed Babel 5 option: base.modules - Use the corresponding module transform plugin in the plugins option. Check out http://babeljs.io/docs/plugins/#modules(…)

### DIFF
--- a/src/transpiler.js
+++ b/src/transpiler.js
@@ -58,7 +58,6 @@ var transpile = (function() {
 
   function babelTranspile(load, babel) {
     var options = this.babelOptions || {};
-    options.modules = 'system';
     if (options.sourceMap === undefined)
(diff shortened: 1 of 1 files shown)


Treri:
I make a [demo](https://github.com/Treri/systemjs-example) to use systemjs and babel to load es6 module directly in browser. The demo will work only when I commen out `options.modules = 'system'`

Treri:
Can anyone review this PR please?

guybedford:
The options.modules is critical to the Babel 5 support in this project. This feature will be deprecated though, and as such does not support Babel 6.

PR #486, opened 2016-05-06
Ensuring that es6 modules are interpreted as es6 modules by today's bundlers/loaders

joeldenning:
See jspm/jspm-cli#1793 for the problem that this fixes. Basically, `Module` objects are not being interpreted as es6 modules by systemjs/builder because they do not define the _esModule property to be true. See [code in the builder](https://github.com/systemjs/builder/blob/master/templates/sfx-core-register.js#L257-L260) for why this is the case.

Interested to know of any problems/concerns anyone has with doing this, but I believe that this is a backwards compatible change unless people are depending on `Module`s *not* being interpreted as es modules.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -34,7 +34,7 @@ module.exports = function(config) {
 
   var files = [
     'test/_helper.js',
-    [options['babel'] ? 'node_modules/regenerator/runtime.js' : ''],
+    [options['babel'] ? 'node_modules/regenerator-runtime/runtime.js' : ''],
 
     [!options.ie8 
         ? (options['babel'] 
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "karma-sauce-launcher": "^0.2.10",
     "minimist": "^1.1.0",
     "mocha": "^2.0.1",
-    "regenerator": "^0.8.9",
+    "regenerator-runtime": "^0.9.5",
     "traceur": "0.0.93",
     "typescript": "^1.6.2"
   },
--- a/src/loader.js
+++ b/src/loader.js
@@ -18,7 +18,11 @@
 *********************************************************************************************
 */
 
-function Module() {}
+function Module() {
+	if (this) {
+		this.__esModule = true;
+	}
+}
 // http://www.ecma-international.org/ecma-262/6.0/#sec-@@tostringtag
 defineProperty(Module.prototype, 'toString', {
   value: function() {
--- a/src/loader.js
+++ b/src/loader.js
@@ -19,7 +19,7 @@
 */
 
 function Module() {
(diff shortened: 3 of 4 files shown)


joeldenning:
Note that the switch from regenerator to regenerator-runtime was necessary because of [a commit to regenerator](https://github.com/facebook/regenerator/commit/18df917126ab854ace16ded77af4164c9b315ea2) that makes the `regenerator/runtime.js` file no longer loadable via a script tag.

joeldenning:
Also I installed this branch of es6-module-loader into systemjs, built systemjs, and then ran the tests for systemjs and that all the tests passed.

guybedford:
This is actually another level of interop that I've deliberately not implemented.

The reason for this is that real ES modules will not be able to support this, unless we want to start advocating all es modules in SystemJS on a native loader do this same export...

My hope here was always to get transpilers to output a check like `module.__esModule || module instanceof Module` kind of check that could be fowards-compatible. I tried to rally es discuss to get a solid mechanism for this a couple of times but wasn't able to. There is still a Babel issue at https://phabricator.babeljs.io/T2148 I created here.

So it comes down to:

1. Land this, and have SystemJS for the next 10 years always put `__esModule` on every module running in the browser.
2. Attempt to walk the hard path of interop involving updating transpilers to outputs like the above.

I'm more tempted to add cruft to the transpiler workflows of today, than the module workflows of tomorrow, hence my resistance here.

guybedford:
What I'd suggest is the following:

* As soon as the `<script type="module">` implementations start landing, it would be worth checking that the `import * as module from 'module'` detection suggested works in all browsers correctly, and post browser bugs where they don't. This way that detection can be prioritised.
* As soon as that detection is known to be reliable, push for implementation of the detection approach in Babel and other transpilers.

If anything goes wrong on the above two steps (eg browser interop breaks for even just one browser hitting a wall), then the above can be implemented. But until then I don't think we should.

joeldenning:
All of that makes sense - one question with it though. You recommended that transpilers check `module instanceof Module`, but Module is wrapped in an iife and thus not globally available to be checked against. Since that is the case, what can transpilers / bundlers do in their interop to check if an object they have is a Module? See the jspm issue linked to in the original post for how SFX builds fail to detect if injected deps are Modules, resulting in a runtime error.

guybedford:
Yes that was as a quick example... the actual detection would be something more like:

```javascript
function isModule(m) {
  return typeof m == 'object' && (m.__esModule || typeof Symbol != 'undefined' && Symbol.toStringTag && m[Symbol.toStringTag] == 'Module');
}
```

Sure, I've responded on that issue... let's continue to track there then on this.

PR #493, opened 2016-06-20
Remove polyfills from 0.50

probins:


src/loader.js
```diff
@@ -696,116 +685,3 @@
     for (var p in descriptors)
       this[p] = descriptors[p];
   }
-
-  function InternalMapPolyfill() {}
```
guybedford: I agree with removing this now, but I think in future we may well recover it in some way as a high-performance polyfill approach for a backwards-compatible loader variant.

guybedford: Sorry @joeldenning... we are not removing the polyfill approach, we're just "hiding" it for now until it is an important implementation concern :)

PR #494, opened 2016-06-20
0.50: change to new globals

probins:


src/declarative.js
```diff
@@ -107,7 +107,7 @@
     return moduleRecords[key] || (moduleRecords[key] = {
       key: key,
       dependencies: [],
-      module: new Module({}),
+      module: new Reflect.Module({}),
```
guybedford: I don't think the above matter too much? `Module` is an internally scoped reference to `Reflect.Module` and not `global.Module` I think?

src/wrapper-end.js
```diff
@@ -3,8 +3,11 @@
     
   var Reflect;
 
-  System = new SystemLoader();
-  System.constructor = SystemLoader;
+  System = System || {};
+  if (!System.loader) {
+    System.loader = new SystemLoader();
+    System.loader.constructor = SystemLoader;
+  }
```
guybedford: I'd be happier with:

```javascript
var loader = new SystemLoader();
loader.constructor = SystemLoader;
```

then in the section below:

```javascript
  if (cjsMode) {
    exports.Loader = Loader;
    exports.Module = Module;
    exports.SystemLoader = SystemLoader;
  }

  __global.LoaderPolyfill = Loader;
  __global.ModulePolyfill = Module;

  __global.Reflect = __global.Reflect || {};
  __global.Reflect.Module = __global.Reflect.Module || Module;
  __global.Reflect.Loader = __global.Reflect.Loader || Loader;

  __global.System = __global.System || {};
  __global.System.global = __global.System.global || __global;
  __global.System.loader = __global.System.loader || loader;
```

src/declarative.js
guybedford: See https://github.com/ModuleLoader/es6-module-loader/blob/0.50/src/loader.js#L682, we rely on internal scoping through the build process.

It may well be worth rewriting the build using ES modules and Rollup as another task here to make these scoping assumptions clearer!

probins: I would prefer having these add-ons as modules :-) `Module()` is surely left over from the old spec, and I think using `Reflect.Module` makes it clearer what's going on here, though I agree it probably doesn't matter much in practice.

probins:
Module() is marked as TODO, so we can leave that till later.

What is System.global used for? It's not in the spec AFAICS.

probins:
the change to travis.yml doesn't seem to have had any effect - does it need to be changed in master too?

src/declarative.js
probins: > It may well be worth rewriting the build using ES modules and Rollup as another task here to make these scoping assumptions clearer!

just to make sure we're on the same wavelength, you're proposing rewriting src/*.js as ES6, and then use Rollup as build instead of concat to convert that into IIFE? I think that would be a very good idea - and would suggest extending to SystemJS as well.

guybedford:
System.global is a separate spec at https://github.com/tc39/proposal-global.

src/declarative.js
guybedford: Yes exactly!

src/wrapper-end.js
probins: @guybedford I don't think this is correct. Tests are failing, and I think this is because cjsMode needs System not SystemLoader

PR #496, opened 2016-06-21
Remove site object from System

probins:
something like this can be put back in as and when the spec gets around to defining it.

### DIFF
--- a/src/system.js
+++ b/src/system.js
@@ -83,59 +83,10 @@
   var SystemLoader = function() {
     Loader.call(this, arguments);
 
-    var siteTable = {};
-    this.site = function(mappings) {
-      for (var m in mappings)
-        siteTable[m] = mappings[m];
-    }
-    this.site.get = function(name) {
-      return siteTable[name];
-    }
-    this.site.set = function(name, url) {
-      siteTable[name] = url;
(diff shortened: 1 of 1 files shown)


probins:
tests are using System.site too. Looks like there's quite a lot of changes needed to the test suite to get it in line with the new spec.

guybedford:
Let's remove the sites tests then? Or perhaps move the implementation of `sites` into a hook in the tests if the tests rely on it too heavily?

probins:
the tests are currently using `site.` for 2 things: in `system.normalize.spec.js`, and to set `babel` and `traceur` locations. As both of these have to be changed anyway (`normalize` no longer exists, and Babel has to change for Babel 6 - if we decide to go with a special build with only import/export transpile, then package.json will have to be changed to reflect that), I don't see much point in doing anything to the tests atm. ISTM the better approach is to decide first what needs to be done to match the polyfill to the spec and the new environment (such as Babel/Traceur), and to ignore the tests until that's done.

probins:
`site.` is also used by `index.js` for setting path for Babel/Traceur

probins:
might as well close this for the moment as well

PR #497, opened 2016-06-21
Limit Travis to master branch

probins:
I think I may be misunderstanding https://docs.travis-ci.com/user/customizing-the-build/#Building-Specific-Branches - depends on whether the branch to ignore is the branch being merged or the branch into which the new branch is being merged. Either way, if we change it to master only, then it should not build any others. As and when we need Travis to run on another branch, we can add it to this whitelist.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,5 +41,5 @@ script:
   - "[ $SAUCE_LABS == false ] || npm run test:browser-$PARSER -- --saucelabs $OPTIONS"
 
 branches:
-  except:
-    - 0.50
+  only:
+    - master


guybedford:
Thinking about this further, I think we should keep the tests failing... Getting the tests in line is a priority.

probins:
see my comment at https://github.com/ModuleLoader/es6-module-loader/pull/496#issuecomment-227687621

probins:
another approach with the tests is to remove them completely from 0.50 to start with, and then gradually add them back in to test the individual functions.

guybedford:
Certainly a lot of the tests relating to these areas can probably be removed, but the main ES semantic tests and custom loader tests are important to keep around and get working with a new approach to replace sites (perhaps just always testing against a custom TestLoader instance that has its own sites alternative).

PR #501, opened 2016-06-22
0.50: Remove hook() from system.js

probins:


guybedford:
In the latest loader, the hooks are set via:

```
Loader.prototype[Reflect.Loader.translate] = ...
```

We need to update properly to the new paradigm, not just delete the old one.

probins:
you're right, I wasn't paying attention to the '@@' in the spec. However, the default `System.loader` is being created in `wrapper-end.js` before `Loader` is set on `Reflect`, so I think either this has to read `this.prototype[Loader.resolve] =...` etc (which doesn't correspond exactly to the spec), or the default `System.loader` will have to be created after `Loader` is set on `Reflect` - or am I missing something?

guybedford:
Yes we don't need to correspond word for word, as long as the right symbol and bindings are coming through (`this.prototype[RESOLVE_SYMBOL] = ...`).

probins:
ok. updated

guybedford:
As you've mentioned as well, I think I would prefer to get the tests running as the first priority, then make compatibility changes between working code states.

probins:
2 test files also use `.hook` so I'll add those to this PR once #507 is merged

PR #504, opened 2016-06-27
Expand on pipeline stages in readme

probins:


guybedford:
Very nicely clarified, thank you!

If you have any suggestions with respect to how to clarify the SystemJS references to the pipeline that would be incredibly welcome as well (although no pressure of course :)).

probins:
yeah, docs in general is something else to add to the to-do list :-)

Issue #506, opened 2016-06-28
Add dist to .gitignore?


probins:
Readme says I shouldn't edit `dist/*` but `dist/*` are needed to run tests, and if I run grunt to generate them they will be included in any new branch uploaded to Github. This probably isn't wanted, so shouldn't dist be added to `.gitignore`?

guybedford:
dist is currently checked in for releases only, and not for commits between releases.

probins:
this has been done for 1.0, so I'll close this

PR #507, opened 2016-06-28
0.50 Update Node environment, and make tests compatible

probins:
This PR:
* updates Node to 4.4+
* reinstates URL polyfill so Node compatible (I see this as a temporary measure until I come up with a better idea ;-))
* couple of other minor updates to package.json
* updates tests to use System.loader
* rejigs wrapper-end so System is exported in cjsMode
* updates regenerator pkg to regenerator-runtime (regenerator can probably be removed)

With this, the main bulk of tests pass, but there are still some issues:
* I've commented out a couple of registry-related tests which don't pass, e.g. because they use the Registry object which is not exported
* for the karma browser tests, Travis uses an ancient version of Chromium, which doesn't support ES6 (I'm not sure how much ES5 it supports!). AIUI this is because Travis uses Ubuntu Precise (4 yrs old) and so the browsers from that era are installed. It also uses an old version of Firefox. It can use the latest Firefox by specifying this in .travis.yml. However, AIUI this can't be done with Chromium, and I think that we would need to specify in .travis.yml to use the Trusty beta, and then the latest versions could be used. For the moment, I have commented out the Chromium browser.
* not sure what's causing the outstanding issues

Despite these outstanding issues, I'd suggest merging this, as it solves the bulk of the test problems.


probins:
and just updated demo as well

.travis.yml
```diff
@@ -13,22 +15,22 @@ env:
     - PARSER="babel"
 matrix:
   include:
-    - node_js: "4.1"
+    - node_js: "4.4"
```
guybedford: Let's just make this "4"?

dist/es6-module-loader-declarative.src.js
```diff
@@ -0,0 +1,1254 @@
+  // from https://gist.github.com/Yaffle/1088850
```
guybedford: Can you please remove the dist diffs from the PR?

guybedford:
If you can please remove the the dist diffs it will make it easier to complete review, but it all looks good to merge.

.travis.yml
probins: that would use the latest, right?

probins:
this was why I raised #506

guybedford:
I'm just going to go ahead and merge this so you don't need to waste any more time given the change of direction here.

probins:
@guybedford given that we're holding the new branch, it might be a good idea to upgrade the environment on the master branch. Things that come to mind are:
* upgrade Node to 4/6
* use latest Firefox on Travis
* consider using Trusty beta on Travis together with latest Chrome
* this should mean the `when` dependency can go
* upgrade to latest Babel/Traceur? Convert to import/export only transpiles?

guybedford:
@probins I think it may be best to look at it as a rewrite at this point actually.

PR #517, opened 2016-10-14
WIP - register dynamic refactoring

guybedford:
The original plan was to implement `System.registerDynamic` (the wrapper format for CommonJS bundling) as a layer on top of this loader in SystemJS, but it has turned out that due to the complexities of the linking algorithm, performance and also integrating the ideas from the latest TC39, it really is best to combine the implementation with `System.register`.

This is the first step of that work here, pending further testing and performance work.


guybedford:
Merged manually.

Issue #527, opened 2017-02-14
When loading commonjs modules, named exports after the default export are 'undefined'


nnajm:
Hello,

Issue type: **Bug**
Version: **2.1.0**

When using systemjs to load transpiled typescript to commonjs modules, if a named export comes after a default export, the named export will be lost.

Example:

- comp.ts
``` typescript
export const b = 3;
export default 1;
export const c = 5;
```
- main.ts
``` typescript
import a, {b, c} from './comp';

console.log(a + b + c); // <== c will be undefined
```

- system.config.js
``` javascript
System.config({
  //use typescript for compilation
  transpiler: 'ts',
  typescriptOptions: {
    "module": "commonjs"
  },
  //map tells the System loader where to look for things
  map: {
    ts: "https://raw.githubusercontent.com/frankwallis/plugin-typescript/master/lib/plugin.js",
    typescript: "https://cdnjs.cloudflare.com/ajax/libs/typescript/2.1.6/typescript.js"
  },
  //packages defines our app package
  packages: {
    src: {
      main: './main'
    }
  },
  meta: {
    "*.ts": {
      "loader": "ts"
    },
    typescript: {
      "exports": "ts"
    }
  }
});
```

**Plunker:** https://plnkr.co/edit/JP20hkNbmryPVYhvcklP?p=preview

---
### Proposed solution

I've found the location of the bug to be in [register-loader.js/doEvaluate](https://github.com/ModuleLoader/es-module-loader/blob/master/core/register-loader.js#L620)

```
// __esModule flag extension support via lifting
if (moduleObj.default && moduleObj.default.__esModule) {
    moduleObj.__useDefault = false;
    for (var p in moduleObj.default) {
        if (Object.hasOwnProperty.call(moduleObj.default, p))
            moduleObj[p] = moduleObj.default[p];
        }
        moduleObj.__esModule = true;
}
```

A solution would be to save a reference to default obj before copying it to moduleObj:

```
var moduleDefault = moduleObj.default;

// __esModule flag extension support via lifting
if (moduleDefault && moduleDefault.__esModule) {
    moduleObj.__useDefault = false;

    for (var p in moduleDefault) {
        if (Object.hasOwnProperty.call(moduleDefault, p))
            moduleObj[p] = moduleDefault[p];
    }

    moduleObj.__esModule = true;
}
```

nnajm:
Created a pull request to fix this bug: #528

guybedford:
[...]

PR #528, opened 2017-02-14
Fix: When loading commonjs modules, named exports after the default export are 'undefined'

nnajm:
A proposed fix for Issue #527.


guybedford:
Perfect, thank you this looks exactly right to me!

